PR #3, opened 2020-03-02
Tweaks to the storage handler

jankoulaga:
Storage handler now:
- respects StorageStrategy = "none" in a way that it can be tweaked to never write to cookie jars, or local storage
- can be overridden by external implementation

The LC Module now accepts the external StorageHandler, and can also return the url for IDEX resolution, so that the module itself doesn't manipulate device storage or doesn't have to issue AJAX calls.


src/live-connect.js
```diff
@@ -88,25 +92,30 @@ export function LiveConnect (liveConnectConfig) {
   }
 
   try {
-    const reducer = (accumulator, func) => accumulator.combineWith(func(accumulator.data))
+    const storageHandler = StorageHandler(configuration.storageStrategy, externalStorageHandler)
```
melgenek: we pass both `configuration` and `externalStorageHandler` in the constructor. Should we wrap the `storage.js` into an object and pass it instead of the `configuration.storageStrategy` if the `configuration.storageStrategy==null`?

src/manager/identifiers.js
```diff
@@ -1,73 +1,113 @@
...
-import { getCookie, getFromLs, getOrAddWithExpiration, setCookie } from '../utils/storage'
-import { isIframe, loadedDomain, parentHostname } from '../utils/page'
+import { loadedDomain } from '../utils/page'
 import { getLegacyId, getLegacyIdentifierKey } from '../utils/legacy'
 import { domainHash } from '../utils/hash'
+import { expiresInDays, strEqualsIgnoreCase } from '../utils/types'
+import { StorageStrategy } from '../model/storage-strategy'
 
-const NEXT_GEN_FP_NAME = '_lc2_duid'
+const NEXT_GEN_FP_NAME = '_lc2_fpi'
```
melgenek: is it ok, that we lose the cookies that are already set?

src/manager/identifiers.js
```diff
@@ -1,73 +1,113 @@
...
 
-/**
- * @return {LegacyId|null|undefined}
- * @private
- */
-function _legacyDuid () {
-  const _legacyEntry = getFromLs(getLegacyIdentifierKey())
-  return getLegacyId(_legacyEntry)
-}
+    const addDays = (days) => new Date().getTime() + (days * 864e5)
```
melgenek: did we try to change the variables with a function value to have only functions instead? Like
```
function addDays(day) {...}
```

It does not seem to be a problem to have nested functions inside the `resolve`.

src/utils/storage.js
```diff
@@ -107,55 +96,28 @@ export function findSimilarInJar (keyLike) {
...
-  try {
-    if (hasLocalStorage()) {
-      ret = _unsafeGetFromLs(key)
-    }
-  } catch (e) {
-    emitter.error('LSFetch', 'Failed fething key from ls', e)
-  }
-  return ret
+export function setCookie (key, value, expires, sameSite, domain) {
+  setCookieInt(key, value, { domain: domain, expires: expires, samesite: sameSite })
```
melgenek: what is the purpose of extracting the `setCookieInt` into a separate method?

src/live-connect.js
jankoulaga: I don't get it... Storage.js is just our implementation  of the StorageHandler, and already fits the signature, so IMHO there's no need to wrap anything there.

src/manager/identifiers.js
jankoulaga: Yes

src/utils/storage.js
jankoulaga: Because the underlying lib expects the JSON with options.

src/manager/identifiers.js
jankoulaga: We don't want functions inside of functions, that's why it's like this.

README.md
```diff
@@ -50,13 +50,42 @@ The object returned after initialisation (`lc` in the snippet above) is exposing
...
+- `resolve` function accepts a callback and an additional object with key value pairs. Of course, errors during resolution will be emitted on the EventBus and sent to the collector. The second parameter is `additionalParameters` which is an object, and will be attached to the IdentityResolution request, split into key-value pairs. The purpose of this object is to include key-value pairs in the request, e.g. for identifiers that cannot be found in the cookie jar, or in LocalStorage, or simply there's a requirement for a certain identifier to be represented under a specific key which doesn't match it's name in the cookie jar, or LocalStorage key.
+- `resolutionCallUrl` function returns the URL to be called in order to receive the resolution to a stable identifier.
+
+### Overriding the StorageHandler.
+LiveConnect can be initialized in a way so that it does not manipulate storage on the device on it's own. For example, if one wants to use it's own handler for storage, it is enough to send the storage handler in the constructor.
+The only thing one needs to adhere to is the signature of each function that's needed on the StorageHandler:
+- `function hasLocalStorage ()`
```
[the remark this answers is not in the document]
jankoulaga: I think they're quite self explanatory.

src/utils/storage.js
melgenek: it is possible to just call
```
cookies.set(key, value, { domain: domain, expires: expires, samesite: sameSite })
```

PR #7, opened 2020-05-25
Optimizations on browserstack tests

jankoulaga:
This PR:
- loads the bundle only once. There's no need to load the rollup bundle from FS every time the tracker is requested
- GZIPs the bundle, to potentially reduce the over the wire transfer between CCI & BS
- introduces retries
- splits the waitUntil into much smaller chunks, speeding up tests, as for cases where `expectedRequests` doesn't match the `server.getHistory().length` doesn't always wait for 4 seconds, but also allows more frequent checks on the condition.

Result: BS tests are down from ~20 mins to ~ 10.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3494,6 +3494,61 @@
         "readable-stream": "^2.3.6"
       }
     },
+    "compressible": {
+      "version": "2.0.18",
+      "resolved": "https://registry.npmjs.org/compressible/-/compressible-2.0.18.tgz",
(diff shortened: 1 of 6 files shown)


dannashirn:
Just curious, why the removal of lambdas?

jankoulaga:
@dannashirn quite easy...

When writing lambdas, the `this` is not in scope any more. So whenever you want access to the mocha context, it's not there.

https://mochajs.org/#arrow-functions

PR #8, opened 2020-06-03
RTA-83. Expose liQ on global window.

jankoulaga:
In order to track events and potentially build segments out of it, liQ interface was exposed.

### DIFF
--- a/src/live-connect.js
+++ b/src/live-connect.js
@@ -41,7 +41,7 @@ import * as errorHandler from './events/error-pixel'
 import * as C from './utils/consts'
 import * as cookies from './enrichers/identifiers'
 import * as legacyDuid from './enrichers/legacy-duid'
-import { isArray, isObject } from './utils/types'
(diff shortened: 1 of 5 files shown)


src/live-connect.js
```diff
@@ -115,11 +135,30 @@ export function LiveConnect (liveConnectConfig, externalStorageHandler) {
...
+/**
+ * @param {LiveConnectConfiguration} liveConnectConfig
+ * @param {StorageHandler} externalStorageHandler
+ * @returns {LiveConnect}
+ * @constructor
+ */
+export function LiveConnect (liveConnectConfig, externalStorageHandler) {
+  console.log('Initializing LiveConnect')
+  try {
+    window && (window.liQ = window.liQ || [])
```
melgenek, line 156: if the `liQ` is an array, should we push the messages into the initialized LiveConnect as we do in Builder https://github.com/LiveIntent/LiveConnectBuilder/blob/master/src/index.js#L75-L82?

PR #24, opened 2020-09-22
RTA-286 Referrer

melgenek:
https://liveintent.atlassian.net/browse/RTA-286

### DIFF
--- a/src/utils/page.js
+++ b/src/utils/page.js
@@ -18,10 +18,13 @@ export function getReferrer (win = window) {
 export function getPage (win = window) {
   const ancestorOrigins = _safeGet(() => win.location.ancestorOrigins) || {}
 
-  const windows = [win]
-  for (let currentWindow = win; currentWindow !== top; currentWindow = currentWindow.parent) {
+  const windows = []
+  let currentWindow = win
+  while (currentWindow !== top) {
     windows.push(currentWindow)
+    currentWindow = currentWindow.parent
   }
+  windows.push(currentWindow)
 
   let detectedPageUrl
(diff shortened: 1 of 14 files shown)


jankoulaga (approved):

src/utils/page.js
```diff
@@ -1,29 +1,46 @@
...
-  return (document.domain || (document.location && document.location.host)) || (window && window.location && window.location.host) || 'localhost'
+export function getPage (win = window) {
+  const ancestorOrigins = _safeGet(() => win.location.ancestorOrigins) || {}
+
+  const windows = []
+  let currentWindow
+  do {
+    currentWindow = currentWindow ? currentWindow.parent : win
+    windows.push(currentWindow)
+  } while (currentWindow !== top)
```
gipeshka, line 26: Might be a bit shorter, if you prefer `for`
```suggestion
  const windows = [win]
  
  for (let currentWindow = win; currentWindow !== top; currentWindow = currentWindow.parent) {
    windows.push(currentWindow)
  }
```

gipeshka (approved):

src/utils/page.js
melgenek: The semantic is different. For loop doesn't keep the window #0, if it's already "top". My code does.

melgenek: `const windows = [win]` I didn't manage to come up with this and my for loop didn't work :facepalm: 

:+1:

jankoulaga:
@melgenek your linting is off ;)

src/utils/page.js
melgenek: this code is wrong

PR #26, opened 2020-09-28
RTA-386 Pluggable storage and ajax

melgenek:


README.md
```diff
@@ -85,7 +91,12 @@ const storageHandler = {
   },
   ...
 }
-const lc = LiveConnect(configOptions, storageHandler)
+const ajaxHandler = {
+  get: (url, responseHandler, fallback, timeout) => {
```
jankoulaga, line 95: Shouldn't it look like this? https://github.com/prebid/Prebid.js/blob/master/src/ajax.js#L18

melgenek: I wanted to adapt that `ajaxBuilder` to our signature. But I can do the other way around.

melgenek: I think providing the ajax that satisfies our signature is a better approach for at least a couple of reasons:
- live-connect needs `withCredentials=true` in ajax, so in prebid we preconfigure the ajax call to always be `withCredentials`
- it's a bit complicated to pass a curried function in js. We would need to check that the outer function is a function-> apply it-> check that the result is a fuction-> call ajax.

gipeshka (approved):
with minor suggestion

src/handlers/call-handler.js
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * @typedef {Object} CallHandler
+ * @property {function} [ajaxGet]
+ * @property {function} [pixelGet]
+ */
+
+const _noOp = () => undefined
+
+/**
+ * @param {CallHandler} externalCallHandler
```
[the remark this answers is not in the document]
melgenek: it does. There is a definition in the top of this file https://github.com/LiveIntent/live-connect/pull/26/files#diff-878cf53bd9649b38c8827ff62abe2a98R4-R8.

src/enrichers/legacy-duid.js
```diff
@@ -9,7 +9,7 @@ export function enrich (state, storageHandler) {
   console.log('legacy-duid.enrich', state)
   const duidLsKey = getLegacyIdentifierKey()
   try {
-    if (state.appId && storageHandler.hasLocalStorage()) {
+    if (state.appId && storageHandler.localStorageIsEnabled()) {
```
jankoulaga, line 12: Omg... :(

src/handlers/call-handler.js
```diff
@@ -0,0 +1,39 @@
...
+ */
+export function CallHandler (externalCallHandler) {
+  const errors = []
+
+  function _externalOrError (functionName) {
+    const hasExternal = externalCallHandler && externalCallHandler[functionName] && isFunction(externalCallHandler[functionName])
+    if (hasExternal) {
+      return externalCallHandler[functionName]
+    } else {
+      errors.push(functionName)
```
jankoulaga, line 25: What's the reason for storing errors here? Isn't it better to just emit those and be done with it?

src/idex/identity-resolver.js
```diff
@@ -74,6 +69,8 @@ export function IdentityResolver (config, storageHandler) {
     const tuples = []
     tuples.push(['duid', encodedOrNull(nonNullConfig.peopleVerifiedId)])
     tuples.push(['us_privacy', encodedOrNull(nonNullConfig.usPrivacyString)])
+    tuples.push(['gdpr', encodedOrNull(nonNullConfig.gdprApplies)])
```
jankoulaga, line 72: Yeah this won't work here... gdpr will be set to false which means it's false and you won't even send that info. Plus, the gdprApplies in the config is boolean, however the string must be 0|1 from what i read.

src/enrichers/legacy-duid.js
melgenek: I did this change to have the same behaviour for the check in prebid and builder.
Prebid's `hasLocalStorage` is returning `!!window.localStorage`, so it just checks that the localStorage exists in the page.
Whereas our implementation of `hasLocalStorage`: puts the value into the ls -> gets the value from ls -> checks that the value was successfully written. This is what `localStorageIsEnabled` does in prebid. 

So the name is changes in our codebase to be `localStorageIsEnabled`.

src/handlers/call-handler.js
melgenek: I just wanted to have only one error message. 
For example, in the storage we expect 7 functions to be provided. And adding errors into the array prevents us from sending 7 errors: one per each missing function.
The implementation is the same in the `call` handler so that there is only 1 error send instead of 2.

src/idex/identity-resolver.js
melgenek: I mixed up the values that are passed into live-connect. I'll update the prebid and live-connect prs.

src/enrichers/legacy-duid.js
jankoulaga: Yes yes! Great find!

src/handlers/call-handler.js
jankoulaga: Gotcha...

src/idex/identity-resolver.js
melgenek: updated https://github.com/LiveIntent/live-connect/pull/26/commits/8e9aa5fe841b8d9be33eb33fa04132d1b10693c7

jankoulaga (approved):

PR #32, opened 2020-11-09
RTA-453. Bundle size optimizations and LC modes

jankoulaga:
This code introduces the following changes:
- using the proces.env to determine which constructor should be used
- moving the btoa into a in-project file, because the original version was creating an overhead of 3k in the bundle
- moving ulid into the code base, as it only includes the needed functionality for constructing the ulid, and there's no deconstruction
- using a simple merge instead of the spread operator
- extracted similar functions from IdexResolver and State


src/pixel/state.js
```diff
@@ -61,70 +61,58 @@ import { toParams } from '../utils/url'
...
-    return _asParamOrEmpty('li_did', dids.join(','), (s) => encodeURIComponent(s))
+    return asStringParam('li_did', dids.join(','))
   },
   hashedEmail: he => {
-    return _asParamOrEmpty('e', he.join(','), (s) => encodeURIComponent(s))
+    return asStringParam('e', he.join(','))
   },
   usPrivacyString: usps => {
-    return _asParamOrEmpty('us_privacy', usps && encodeURIComponent(usps), (s) => encodeURIComponent(s))
+    return asStringParam('us_privacy', usps && encodeURIComponent(usps))
```
melgenek, line 103:
```suggestion
    return asStringParam('us_privacy', usps)
```
why are we passing the value as `encodeURIComponent(value)` and then transforming it once more with `encodeURIComponent` inside the `asStringParam`?
After removing all the ` && encodeURIComponent` here, the unit and integration tests pass.

test/unit/events/replayemitter.spec.js
```diff
@@ -2,108 +2,108 @@ import { expect, spy, use } from 'chai'
 import chaiSpies from 'chai-spies'
 import ReplayEmitter from '../../../src/events/replayemitter'
 
-use(chaiSpies);
+use(chaiSpies)
```
melgenek, line 5: should we update the `package.json` to run lint on the tests? The Builder is already doing the test linting https://github.com/LiveIntent/LiveConnectBuilder/pull/1/commits/101b416b54bd86d296a7588db57e1d688096966d

src/minimal-live-connect.js
```diff
@@ -0,0 +1,85 @@
...
+ * @property {(number|undefined)} ajaxTimeout
+ * @property {(string|undefined)} source
+ * @property {(number|undefined)} publisherId
+ * @property {(string|undefined)} url
+ */
+
+/**
+ * @typedef {Object} LiveConnectConfiguration
+ * @property {(string|undefined)} appId
+ * @property {(StorageStrategy|null)} storageStrategy
```
melgenek, line 24:
```suggestion
 * @property {(string|null)} storageStrategy
```

src/handlers/read-storage-handler.js
```diff
@@ -0,0 +1,47 @@
+import { isFunction, strEqualsIgnoreCase } from '../utils/types'
+import { StorageStrategy } from '../model/storage-strategy'
+import * as emitter from '../utils/emitter'
```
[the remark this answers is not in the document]
jankoulaga: And what should we do with that? The minimal LC isn't even supposed to send any pixels, however, the read only might be used wherever the emitter is present. I will add tests for this.

src/idex/identity-resolver-nocache.js
```diff
@@ -0,0 +1,79 @@
+import { toParams } from '../utils/url'
```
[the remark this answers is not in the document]
jankoulaga: I didn't want to make that decision in this step.

src/pixel/state.js
jankoulaga: True...

melgenek (approved):

test/unit/events/replayemitter.spec.js
jankoulaga: Yes, and no... however if i run lining on specs, then i'm getting a lot of complaints about describes and its.
I think that's the reason why it's not enabled in the first place

PR #38, opened 2021-04-06
CM-18 - Set referrer policy

3link:
It is a quick-fix. We need more research. 

It looks like we hit the issue because of the behaviour described here: https://developers.google.com/web/updates/2020/07/referrer-policy-new-chrome-default.

**Open questions:**
- [x] The behaviour change was planned for Chrome version 85, we are now running tests with 89. Why did we see the issue only now?
- [x] Do we need to inform our customers/adjust any integrations?
- [x] It would be good to have some research on how the privacy behaviour will develop over time. Do we need a long term plan?

### DIFF
--- a/COLLECTOR_PARAMS.md
+++ b/COLLECTOR_PARAMS.md
@@ -5,7 +5,7 @@
 ### `se`
 - contains the b64 url encoded string of the JSON that was sent via `liveConnect.push` function
 ### `duid`
-- contains the LiveConnect managed first party identifier, in the `${apexDomainHash}--${ULID}`, or `${iframeDomainHash}-${websiteUrlDomainHash}--${ULID}` format
+- contains the LiveConnect managed first party identifier, in the `${apexDomainHash}--${ULID}`
 ### `tna`
 - contains the `config.trackerName`
 ### `pu`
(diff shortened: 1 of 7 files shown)


COLLECTOR_PARAMS.md
```diff
@@ -5,7 +5,7 @@
 ### `se`
 - contains the b64 url encoded string of the JSON that was sent via `liveConnect.push` function
 ### `duid`
-- contains the LiveConnect managed first party identifier, in the `${apexDomainHash}--${ULID}`, or `${iframeDomainHash}-${websiteUrlDomainHash}--${ULID}` format
```
3link, line 8: Our of interest: why did the `${iframeDomainHash}-${websiteUrlDomainHash}--${ULID}` part go?

melgenek: Imagine that there's an iframe in two pages:
```
nike.com
 --> googletackingbla.com
     --> our js

adidas.com
 --> googletackingbla.com
     --> our js
```
In the old setup, the cookies are `nike.com--googletackingbla.com--theid1` and `adidas.com--googletackingbla.com--theid2`. The cookies are different for the same user. Though it's possible to have a cross-site cookie via the iframe `googletackingbla.com--theid`, cause the iframe url is always the same. 

This is my understanding.

melgenek (approved):

paulosimsk:
Closing this one, for https://github.com/LiveIntent/live-connect/pull/39

PR #42, opened 2021-06-08
CM-75: update timeout for tests

paulosimsk:
Short description if any.

Author Todo List:

- [x] Build in CI passes
- [x] Self-Review

### DIFF
--- a/test-config/wdio.browserstack.conf.js
+++ b/test-config/wdio.browserstack.conf.js
@@ -118,7 +118,7 @@ exports.config = {
   baseUrl: 'http://localhost',
   //
   // Default timeout for all waitFor* commands.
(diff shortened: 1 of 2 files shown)


paulosimsk:
I noted a couple of reasons for flakiness on tests.

- Tests run specially on MicrosoftEdge (v10) on WINDOWS are more prone to timeout. This is solved increasing the timeouts.
- Browserstack server slow responsiveness. This is a more rare condition that I could not and in such cases we hit the hard limit `waitforTimeout: 90000`
- It seems also tests might break when we try running concurrently Browserstack tests from 2 different places (circleci and local machine for example).

wi101 (approved):

mschuwalow (approved):

PR #46, opened 2021-08-03
Add the --disable-gpu option

3link:
Going from this [version](https://hub.docker.com/layers/selenium/standalone-chrome/3.141.59-20210713/images/sha256-bc4023992691ab8e2f20297f334fdcddd982928fbd969239b39b3dbc2dfc0657?context=explore) to this [version](https://hub.docker.com/layers/selenium/standalone-chrome/3.141.59-20210729/images/sha256-9943ba9d2a89e984080f5681b390229fb47f7d3bb732e4f440456fc52335bae8?context=explore) of standalone Chrome our tests broke. The reason and solution are described here: https://github.com/SeleniumHQ/docker-selenium/issues/1346. 

AFAICS, we may leave the option there forever, as it does not influence our tests.

Author Todo List:

- [x] Build in CI passes
- [x] Self-Review

### DIFF
--- a/test-config/wdio.docker.chrome.conf.js
+++ b/test-config/wdio.docker.chrome.conf.js
@@ -50,7 +50,7 @@ exports.config = {
   //
   capabilities: [{
     browserName: 'chrome',
-    chromeOptions: {
+    'goog:chromeOptions': {
       args: ['--disable-gpu']
     }
   }],
--- a/test-config/wdio.docker.chrome.conf.js
+++ b/test-config/wdio.docker.chrome.conf.js
@@ -49,7 +49,10 @@ exports.config = {
   // https://docs.saucelabs.com/reference/platforms-configurator
   //
   capabilities: [{
-    browserName: 'chrome'
+    browserName: 'chrome',
+    chromeOptions: {
+      args: ['--disable-gpu']
+    }
   }],
   //
   // ===================
@@ -100,7 +103,7 @@ exports.config = {
   // commands. Instead, they hook themselves up into the test process.
   services: ['docker'],
   dockerOptions: {
-    image: 'selenium/standalone-chrome:3.141.59-20210713',
+    image: 'selenium/standalone-chrome:3.141.59',
     healthCheck: 'http://bln.test.liveintent.com:4444',
     options: {
       p: ['4444:4444'],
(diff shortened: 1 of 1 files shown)


melgenek:
let's update the image version to use the specific 3.141.59 chrome rather than the latest https://github.com/LiveIntent/live-connect/pull/46/files#diff-40f4286232c70a210194c5e2b559f1906ee0287effb859504c440bc6571a4558R106

3link:
> let's update the image version to use the specific 3.141.59 chrome rather than the latest https://github.com/LiveIntent/live-connect/pull/46/files#diff-40f4286232c70a210194c5e2b559f1906ee0287effb859504c440bc6571a4558R106

We could. [This commit](https://github.com/LiveIntent/live-connect/pull/46/commits/f036f0b4fd47d26ac751321b69088e409ae2d15e) did it. Do you see a disadvantage in using the  --disable-gpu option?

melgenek:
I don't see a disadvantage in using the "--disable-gpu" option. I don't want to deal with unexpected changes that can appear in chrome in the future. Using the "latest" is usually not the best idea.

3link:
> I don't see a disadvantage in using the "--disable-gpu" option. I don't want to deal with unexpected changes that can appear in chrome in the future. Using the "latest" is usually not the best idea.

It was useful in the past. 

A question related to that: do you know why we do not run the browserstack tests against the latest stable version of each of the browsers?

melgenek:
> It was useful in the past.
there was no intention behind using this version of selenium. We didn't expect the chrome version to change.

> A question related to that: do you know why we do not run the browserstack tests against the latest stable version of each of the browsers?
No reason

3link:
> > It was useful in the past.
> > there was no intention behind using this version of selenium. We didn't expect the chrome version to change.
> 
> > A question related to that: do you know why we do not run the browserstack tests against the latest stable version of each of the browsers?
> > No reason

What is the goal of the simple test? Is it rather a smoke test before tests with browserstack are executed?

melgenek:
> What is the goal of the simple test? Is it rather a smoke test before tests with browserstack are executed?

Browsertest takes a lot of time. The docker test is enough to check the behaviour of the test in normal conditions, and then proceed to checking if the logic works in cross-browser conditions.

3link:
@melgenek and I had a call. We agreed to leave the PR as is and test with the latest Chrome version. Additionally, browserstack tests will be adjusted to also test against latest versions.

paulosimsk:
> @melgenek and I had a call. We agreed to leave the PR as is and test with the latest Chrome version. Additionally, browserstack tests will be adjusted to also test against latest versions.

Sounds good to me

paulosimsk (approved):

PR #47, opened 2021-08-04
Update browser versions list 

3link:
Short description if any.

Author Todo List:

- [ ] Build in CI passes
- [ ] Self-Review

### DIFF
--- a/test-config/wdio.browserstack.conf.js
+++ b/test-config/wdio.browserstack.conf.js
@@ -10,26 +10,25 @@ const commonCapabilities = {
 }
 
 const allCapabilities = [
-  { ...commonCapabilities, browserName: 'Chrome', browser_version: '89.0', os: 'Windows', os_version: '10' },
-  { ...commonCapabilities, browserName: 'Chrome', browser_version: '74.0', os: 'Windows', os_version: '10' },
+  { ...commonCapabilities, browserName: 'Chrome', browser_version: 'latest', os: 'Windows', os_version: '10' },
   { ...commonCapabilities, browserName: 'Chrome', browser_version: '40.0', os: 'Windows', os_version: '7' },
-  { ...commonCapabilities, browserName: 'firefox', browser_version: '87.0', os: 'Windows', os_version: '10' },
-  { ...commonCapabilities, browserName: 'firefox', browser_version: '67.0', os: 'Windows', os_version: '10' },
+  { ...commonCapabilities, browserName: 'firefox', browser_version: 'latest', os: 'Windows', os_version: '10' },
   { ...commonCapabilities, browserName: 'firefox', browser_version: '52.0', os: 'Windows', os_version: '7' },
+  { ...commonCapabilities, browserName: 'Safari', browser_version: 'latest', os: 'OS X', os_version: 'Big Sur' },
   { ...commonCapabilities, browserName: 'Safari', browser_version: '14.0', os: 'OS X', os_version: 'Big Sur' },
   { ...commonCapabilities, browserName: 'Safari', browser_version: '13.1', os: 'OS X', os_version: 'Catalina' },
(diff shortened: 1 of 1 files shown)


test-config/wdio.browserstack.conf.js
```diff
@@ -10,26 +10,25 @@ const commonCapabilities = {
...
   { ...commonCapabilities, browserName: 'Safari', browser_version: '12.0', os: 'OS X', os_version: 'Mojave' },
   { ...commonCapabilities, browserName: 'Safari', browser_version: '11.1', os: 'OS X', os_version: 'High Sierra' },
   { ...commonCapabilities, browserName: 'Safari', browser_version: '10.1', os: 'OS X', os_version: 'Sierra' },
-  { ...commonCapabilities, browserName: 'Safari', browser_version: '7.1', os: 'OS X', os_version: 'Mavericks' },
   { ...commonCapabilities, browserName: 'IE', browser_version: '11.0', os: 'Windows', os_version: '7' },
-  { ...commonCapabilities, browserName: 'IE', browser_version: '10.0', os: 'Windows', os_version: '7' },
-  { ...commonCapabilities, browserName: 'IE', browser_version: '9.0', os: 'Windows', os_version: '7' },
-  { ...commonCapabilities, browserName: 'Edge', browser_version: '18.0', os: 'Windows', os_version: '10' },
+  { ...commonCapabilities, browserName: 'Edge', browser_version: 'latest', os: 'Windows', os_version: '10' },
   { ...commonCapabilities, browserName: 'Edge', browser_version: '15.0', os: 'Windows', os_version: '10', 'browserstack.selenium_version': '3.141.59' },
```
3link, line 25: Do we need to pin the version here?

melgenek: it was a hack to make it work. So yes, the selenium version is intentionally specified here.

Thought, I would delete the Edge 15 and 18 and keep only the latest version. The 15 and 18 don't seem to receive any traffic (out of top 1000 browsers)

test-config/wdio.browserstack.conf.js
```diff
@@ -10,26 +10,25 @@ const commonCapabilities = {
...
+  { ...commonCapabilities, browserName: 'Safari', browser_version: 'latest', os_version: '14', device: 'iPhone 12', real_mobile: true },
+  { ...commonCapabilities, browserName: 'Safari', browser_version: '14.0', os_version: '13', device: 'iPhone 12', real_mobile: true },
+  { ...commonCapabilities, browserName: 'Safari', browser_version: '13.0', os_version: '12', device: 'iPhone 8', real_mobile: true },
   { ...commonCapabilities, browserName: 'Safari', browser_version: '12.0', os_version: '12', device: 'iPhone 8', real_mobile: true },
   { ...commonCapabilities, browserName: 'Safari', browser_version: '11.0', os_version: '11', device: 'iPad Pro 9.7 2016', real_mobile: true },
-  { ...commonCapabilities, browserName: 'Chrome', browser_version: '73.0', os_version: '7.0', device: 'Samsung Galaxy S8', real_mobile: true },
+  { ...commonCapabilities, browserName: 'Chrome', browser_version: 'latest', os_version: '8.0', device: 'Samsung Galaxy S10', real_mobile: true },
```
melgenek, line 31: is it possible to make the device and os_version "latest" as well as the browser version? The thing is that Browserstack has some specific browsers per device. And even with the "latest" browser version we will not run tests for the newly released browsers, cause they are just not available for the device.

3link: Will test it, so far I picket the latest version of the device manually, the first run does nor look good, though: https://automate.browserstack.com/dashboard/v2/builds/4a1e97a55c5756ba5abc647307381aa1454afced

3link: At least providing latest as device seems not be documented anywhere.

PR #49, opened 2021-09-24
Update browser versions

melgenek:
https://liveintent.atlassian.net/browse/CM-197

Author Todo List:

- [X] Add/adjust tests (if applicable)
- [X] Build in CI passes
- [X] Latest master revision is merged into the branch
- [X] Self-Review

### DIFF
--- a/test-config/wdio.browserstack.conf.js
+++ b/test-config/wdio.browserstack.conf.js
@@ -9,27 +9,25 @@ const commonCapabilities = {
   'browserstack.appium_version': '1.14.0'
 }
 
+// https://www.browserstack.com/automate/capabilities
 const allCapabilities = [
-  { ...commonCapabilities, browserName: 'Chrome', browser_version: '89.0', os: 'Windows', os_version: '10' },
-  { ...commonCapabilities, browserName: 'Chrome', browser_version: '74.0', os: 'Windows', os_version: '10' },
+  { ...commonCapabilities, browserName: 'Chrome', browser_version: 'latest', os: 'Windows', os_version: '10' },
   { ...commonCapabilities, browserName: 'Chrome', browser_version: '40.0', os: 'Windows', os_version: '7' },
-  { ...commonCapabilities, browserName: 'firefox', browser_version: '87.0', os: 'Windows', os_version: '10' },
-  { ...commonCapabilities, browserName: 'firefox', browser_version: '67.0', os: 'Windows', os_version: '10' },
+  { ...commonCapabilities, browserName: 'firefox', browser_version: 'latest', os: 'Windows', os_version: '10' },
   { ...commonCapabilities, browserName: 'firefox', browser_version: '52.0', os: 'Windows', os_version: '7' },
-  { ...commonCapabilities, browserName: 'Safari', browser_version: '14.0', os: 'OS X', os_version: 'Big Sur' },
+  { ...commonCapabilities, browserName: 'Safari', browser_version: '14.1', os: 'OS X', os_version: 'Big Sur' },
   { ...commonCapabilities, browserName: 'Safari', browser_version: '13.1', os: 'OS X', os_version: 'Catalina' },
-  { ...commonCapabilities, browserName: 'Safari', browser_version: '12.0', os: 'OS X', os_version: 'Mojave' },
+  { ...commonCapabilities, browserName: 'Safari', browser_version: '12.1', os: 'OS X', os_version: 'Mojave' },
   { ...commonCapabilities, browserName: 'Safari', browser_version: '11.1', os: 'OS X', os_version: 'High Sierra' },
   { ...commonCapabilities, browserName: 'Safari', browser_version: '10.1', os: 'OS X', os_version: 'Sierra' },
-  { ...commonCapabilities, browserName: 'Safari', browser_version: '7.1', os: 'OS X', os_version: 'Mavericks' },
-  { ...commonCapabilities, browserName: 'IE', browser_version: '11.0', os: 'Windows', os_version: '7' },
-  { ...commonCapabilities, browserName: 'IE', browser_version: '10.0', os: 'Windows', os_version: '7' },
-  { ...commonCapabilities, browserName: 'IE', browser_version: '9.0', os: 'Windows', os_version: '7' },
-  { ...commonCapabilities, browserName: 'Edge', browser_version: '18.0', os: 'Windows', os_version: '10' },
-  { ...commonCapabilities, browserName: 'Edge', browser_version: '15.0', os: 'Windows', os_version: '10', 'browserstack.selenium_version': '3.141.59' },
+  { ...commonCapabilities, browserName: 'IE', browser_version: '11.0', os: 'Windows', os_version: '10' },
+  { ...commonCapabilities, browserName: 'Edge', browser_version: 'latest', os: 'Windows', os_version: '10', 'browserstack.use_w3c': true },
+  { ...commonCapabilities, browserName: 'Safari', browser_version: '15.0', os_version: '15 Beta', device: 'iPhone 11 Pro', real_mobile: true },
+  { ...commonCapabilities, browserName: 'Safari', browser_version: '14.0', os_version: '14', device: 'iPhone 12', real_mobile: true },
+  { ...commonCapabilities, browserName: 'Safari', browser_version: '13.0', os_version: '13', device: 'iPhone 11', real_mobile: true },
   { ...commonCapabilities, browserName: 'Safari', browser_version: '12.0', os_version: '12', device: 'iPhone 8', real_mobile: true },
   { ...commonCapabilities, browserName: 'Safari', browser_version: '11.0', os_version: '11', device: 'iPad Pro 9.7 2016', real_mobile: true },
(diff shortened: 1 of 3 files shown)


test-config/wdio.browserstack.conf.js
```diff
@@ -9,27 +9,25 @@ const commonCapabilities = {
...
   { ...commonCapabilities, browserName: 'Safari', browser_version: '11.1', os: 'OS X', os_version: 'High Sierra' },
   { ...commonCapabilities, browserName: 'Safari', browser_version: '10.1', os: 'OS X', os_version: 'Sierra' },
-  { ...commonCapabilities, browserName: 'Safari', browser_version: '7.1', os: 'OS X', os_version: 'Mavericks' },
-  { ...commonCapabilities, browserName: 'IE', browser_version: '11.0', os: 'Windows', os_version: '7' },
-  { ...commonCapabilities, browserName: 'IE', browser_version: '10.0', os: 'Windows', os_version: '7' },
-  { ...commonCapabilities, browserName: 'IE', browser_version: '9.0', os: 'Windows', os_version: '7' },
-  { ...commonCapabilities, browserName: 'Edge', browser_version: '18.0', os: 'Windows', os_version: '10' },
-  { ...commonCapabilities, browserName: 'Edge', browser_version: '15.0', os: 'Windows', os_version: '10', 'browserstack.selenium_version': '3.141.59' },
+  { ...commonCapabilities, browserName: 'IE', browser_version: '11.0', os: 'Windows', os_version: '10' },
+  { ...commonCapabilities, browserName: 'Edge', browser_version: 'latest', os: 'Windows', os_version: '10', 'browserstack.use_w3c': true },
```
melgenek, line 24: the edge doesn't work with the "legacy" protocol, so the "w3c" is enabled. https://github.com/webdriverio/webdriverio/issues/3802
I didn't enable it for all the browsers, cause it fails in the older iphones.

test-config/wdio.browserstack.conf.js
```diff
@@ -9,27 +9,25 @@ const commonCapabilities = {
...
-  { ...commonCapabilities, browserName: 'Chrome', browser_version: '89.0', os: 'Windows', os_version: '10' },
-  { ...commonCapabilities, browserName: 'Chrome', browser_version: '74.0', os: 'Windows', os_version: '10' },
+  { ...commonCapabilities, browserName: 'Chrome', browser_version: 'latest', os: 'Windows', os_version: '10' },
   { ...commonCapabilities, browserName: 'Chrome', browser_version: '40.0', os: 'Windows', os_version: '7' },
-  { ...commonCapabilities, browserName: 'firefox', browser_version: '87.0', os: 'Windows', os_version: '10' },
-  { ...commonCapabilities, browserName: 'firefox', browser_version: '67.0', os: 'Windows', os_version: '10' },
+  { ...commonCapabilities, browserName: 'firefox', browser_version: 'latest', os: 'Windows', os_version: '10' },
   { ...commonCapabilities, browserName: 'firefox', browser_version: '52.0', os: 'Windows', os_version: '7' },
-  { ...commonCapabilities, browserName: 'Safari', browser_version: '14.0', os: 'OS X', os_version: 'Big Sur' },
+  { ...commonCapabilities, browserName: 'Safari', browser_version: '14.1', os: 'OS X', os_version: 'Big Sur' },
```
melgenek, line 18: the Mac and iPhone mostly have 1 browser per os version (`.0` and `.1` versions actually). I prefer to show them explicitly rather than using the `latest`. `latest` in this situation would always be one version anyway.

test-config/wdio.browserstack.conf.js
```diff
@@ -9,27 +9,25 @@ const commonCapabilities = {
   'browserstack.appium_version': '1.14.0'
 }
 
+// https://www.browserstack.com/automate/capabilities
 const allCapabilities = [
-  { ...commonCapabilities, browserName: 'Chrome', browser_version: '89.0', os: 'Windows', os_version: '10' },
-  { ...commonCapabilities, browserName: 'Chrome', browser_version: '74.0', os: 'Windows', os_version: '10' },
+  { ...commonCapabilities, browserName: 'Chrome', browser_version: 'latest', os: 'Windows', os_version: '10' },
```
melgenek, line 14: there's no way to set the `latest` os version, so only the browser version is latest.

test-config/wdio.browserstack.conf.js
```diff
@@ -9,27 +9,25 @@ const commonCapabilities = {
...
   { ...commonCapabilities, browserName: 'Safari', browser_version: '10.1', os: 'OS X', os_version: 'Sierra' },
-  { ...commonCapabilities, browserName: 'Safari', browser_version: '7.1', os: 'OS X', os_version: 'Mavericks' },
-  { ...commonCapabilities, browserName: 'IE', browser_version: '11.0', os: 'Windows', os_version: '7' },
-  { ...commonCapabilities, browserName: 'IE', browser_version: '10.0', os: 'Windows', os_version: '7' },
-  { ...commonCapabilities, browserName: 'IE', browser_version: '9.0', os: 'Windows', os_version: '7' },
-  { ...commonCapabilities, browserName: 'Edge', browser_version: '18.0', os: 'Windows', os_version: '10' },
-  { ...commonCapabilities, browserName: 'Edge', browser_version: '15.0', os: 'Windows', os_version: '10', 'browserstack.selenium_version': '3.141.59' },
+  { ...commonCapabilities, browserName: 'IE', browser_version: '11.0', os: 'Windows', os_version: '10' },
+  { ...commonCapabilities, browserName: 'Edge', browser_version: 'latest', os: 'Windows', os_version: '10', 'browserstack.use_w3c': true },
+  { ...commonCapabilities, browserName: 'Safari', browser_version: '15.0', os_version: '15 Beta', device: 'iPhone 11 Pro', real_mobile: true },
```
melgenek, line 25: I decided to add ios 15 cause everybody is talking about it. It is only in beta in Browserstack for now.

test-config/wdio.browserstack.conf.js
3link: For `Chrome` and `Firefox` we could also test `latest - 1` to make sure changes done to support next lates still work with the previous version.

test/it/live-connect.spec.js
```diff
@@ -189,7 +189,9 @@ describe('LiveConnect', function () {
     waitForRequests(probeLS() ? 1 : 2, server)
 
     const firstTrackingRequest = server.getTrackingRequests()[0]
-    if (!isSafari71() && !isIE9Or10()) {
+    if (isMobileSafariAfter14()) {
+      expect(firstTrackingRequest.query.refr).to.match(/http:\/\/schmoogle.com:3001.*/)
```
3link, line 193: What is actually matched here with .* here?

3link (changes requested):
The test for `Firefox 52/Win 7` seems to fail: https://automate.browserstack.com/dashboard/v2/builds/58fe471cfd7f2f6e87a9669a472f98e860253958/sessions/06bbb2d896154c7646ecfcd74a51781270219679. Is our reporting broken?

melgenek:
> The test for Firefox 52/Win 7 seems to fail: https://automate.browserstack.com/dashboard/v2/builds/58fe471cfd7f2f6e87a9669a472f98e860253958/sessions/06bbb2d896154c7646ecfcd74a51781270219679. Is our reporting broken?

Browserstack doesn't know about the test retries. So that specific test run failed, but it succeeded after the browser restart.

test/it/live-connect.spec.js
melgenek: browserstack ios 15 adds a `/` at the end. The regex could be `http:\/\/schmoogle.com:3001\/?`

test-config/wdio.browserstack.conf.js
melgenek: not sure that there's any value in this. Let's say chrome 88 and 89, doesn't seem to be a big difference. The breaking changes usually come in specific releases.

3link: I was thinking about the cases where we would need to change the behaviour of live-connect because of changes in the latest browser version. In the current set-up we would test this change against the latest browser version and a very old one - for Chrome and Firefox. This might be enough. On the other side, in my naive assumption, I would say we care about `latest` and `latest - 1` the most, because browser adoption is fast nowadays. But there is no guarantee that we can derive "this works for latest - 1" from looking at either of the versions.

test/it/live-connect.spec.js
3link: Do we need `?` if `/` is always added for `isMobileSafariAfter14()`?

test-config/wdio.browserstack.conf.js
melgenek: I would rather explicitly set the versions that we would like to be tested.

test/it/live-connect.spec.js
3link: Then `isMobileSafariAfter14` is misleading. Should it rather be something like `isMobileSafari14OrNewer`?

melgenek: ok https://github.com/LiveIntent/live-connect/pull/49/commits/19a6a511a61e051d224677cba3ad8409802ea974

3link (approved):

PR #50, opened 2021-09-29
Navy 117/collect data

paulosimsk:
Short description if any.

Author Todo List:

- [ ] Build in CI passes
- [ ] Self-Review

### DIFF
--- a/src/utils/consts.js
+++ b/src/utils/consts.js
@@ -8,3 +8,4 @@ export const PEOPLE_VERIFIED_LS_ENTRY = '_li_duid'
 export const DEFAULT_IDEX_EXPIRATION_HOURS = 1
 export const DEFAULT_IDEX_AJAX_TIMEOUT = 5000
 export const DEFAULT_IDEX_URL = 'https://idx.liadm.com/idex'
+export const DEFAULT_CONTEXT_ELEMENT_LENGTH = 5000
--- a/src/utils/email.js
+++ b/src/utils/email.js
@@ -12,6 +12,7 @@ export function isEmail (s) {
 }
 
 const emailLikeRegex = /"([^"]+(@|%40)[^"]+[.][a-z]*(\s+)?)(\\"|")/
+export const emailLikeNoDoubleQuotesRegex = /([a-zA-Z0-9._-]+@[a-zA-Z0-9._-]+\.[a-zA-Z0-9_-]+)/
 
 /**
  * @param {string} s
@@ -30,9 +31,9 @@ export function extractEmail (s) {
  * @param {string} s
  * @returns {string[]}
  */
-export function listEmailsInString (s) {
+export function listEmailsInString (s, regex) {
   const result = []
-  const multipleEmailLikeRegex = new RegExp(emailLikeRegex.source, 'g')
+  const multipleEmailLikeRegex = new RegExp(regex.source, 'g')
   let current = multipleEmailLikeRegex.exec(s)
   while (current) {
     result.push(trim(current[1]))
@@ -47,7 +48,7 @@ export function listEmailsInString (s) {
  */
 export function findAndReplaceRawEmails (originalString) {
   if (containsEmailField(originalString)) {
-    return _replaceEmailsWithHashes(originalString)
+    return replaceEmailsWithHashes(originalString, emailLikeRegex)
   } else if (isEmail(originalString)) {
     const hashes = hashEmail(originalString)
     return {
@@ -66,10 +67,9 @@ export function findAndReplaceRawEmails (originalString) {
  *
  * @param originalString
  * @returns {{hashesFromOriginalString: HashedEmail[], stringWithoutRawEmails: string}}
- * @private
  */
-function _replaceEmailsWithHashes (originalString) {
-  const emailsInString = listEmailsInString(originalString)
+export function replaceEmailsWithHashes (originalString, regex) {
+  const emailsInString = listEmailsInString(originalString, regex)
   const hashes = []
   for (let i = 0; i < emailsInString.length; i++) {
     const email = emailsInString[i]
--- a/src/utils/email.js
+++ b/src/utils/email.js
@@ -1,4 +1,5 @@
 import { trim } from './types'
+import { hashEmail } from '../utils/hash'
 
 const emailRegex = () => /\S+(@|%40)\S+\.\S+/
 
@@ -38,4 +39,46 @@ export function listEmailsInString (s) {
     current = multipleEmailLikeRegex.exec(s)
   }
   return result
+}
+
+/**
+ * @param {string} originalString
+ * @returns {{hashesFromOriginalString: HashedEmail[], stringWithoutRawEmails: string}}
+ */
+export function findAndReplaceRawEmails (originalString) {
+  if (containsEmailField(originalString)) {
+    return _replaceEmailsWithHashes(originalString)
+  } else if (isEmail(originalString)) {
+    const hashes = hashEmail(originalString)
+    return {
+      stringWithoutRawEmails: hashes.md5,
+      hashesFromOriginalString: [hashes]
+    }
+  } else {
+    return {
+      stringWithoutRawEmails: originalString,
+      hashesFromOriginalString: []
+    }
+  }
+}
+
+/**
+ *
+ * @param originalString
+ * @returns {{hashesFromOriginalString: HashedEmail[], stringWithoutRawEmails: string}}
+ * @private
+ */
+function _replaceEmailsWithHashes (originalString) {
+  const emailsInString = listEmailsInString(originalString)
+  const hashes = []
+  for (let i = 0; i < emailsInString.length; i++) {
+    const email = emailsInString[i]
+    const emailHashes = hashEmail(email)
+    originalString = originalString.replace(email, emailHashes.md5)
+    hashes.push(emailHashes)
+  }
+  return {
+    stringWithoutRawEmails: originalString,
+    hashesFromOriginalString: hashes
+  }
 }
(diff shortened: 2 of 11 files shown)


src/utils/email.js
```diff
@@ -11,6 +12,7 @@ export function isEmail (s) {
 }
 
 const emailLikeRegex = /"([^"]+(@|%40)[^"]+[.][a-z]*(\s+)?)(\\"|")/
+export const emailLikeNoDoubleQuotesRegex = /([a-zA-Z0-9._-]+@[a-zA-Z0-9._-]+\.[a-zA-Z0-9_-]+)/
```
3link, line 15: I think [a-zA-Z-0-9] is too restrictive since it matches ASCII only and therefore would miss many valid emails. Usually (.+)@(\S+) is used. Maybe we can go with something like (?i)(?:mailto:)?(.+)@(\S+)? Not sure if modifiers and non-capturing groups are supported here, though.

src/utils/page.js
```diff
@@ -38,6 +41,27 @@ export function getPage (win = window) {
...
 
+/**
+ * @return {string|undefined}
+ */
+export function getContextElements (contextSelectors, contextElementsLength, win = window) {
+  if (!contextSelectors || contextSelectors === '') {
+    return ''
+  } else {
+    var collectedElements = _collectElementsText(contextSelectors, win)
+    return base64UrlEncode(collectedElements).slice(0, contextElementsLength)
```
3link, line 52: This might produce an encoding that is considered invalid by some base64 parsers. Since base64 encodes a byte triplet to a byte quadruple, we could slice from 0 to contextElementsLength - contextElementsLength % 4 to produce a correct encoding.

3link, line 52: We also might think about only collecting as much data as we would eventually send by checking `4 * ceil(n/3.0) < contextElementsLength` where `n` is the number of bits in the utf-8 encoded accumulator for the elements added to it so far. It is debatable, but might prevent building huge strings that we would cut off either way.

test/unit/enricher/page.spec.js
```diff
@@ -13,12 +13,26 @@ describe('PageEnricher', () => {
...
   })
 
-  it('should return the url and the referrer of the page"', function () {
-    const state = {}
+  it('should return the url, referrer and the contextElements of the page', function () {
+    var newHeadline = document.createElement("h1")
+    var content = document.createTextNode("Some header")
+    newHeadline.appendChild(content)
+    var newP = document.createElement("p")
+    var newContentEmail = document.createTextNode("mailTo: john@test.com, also found: another@test.com !")
```
3link, line 21: AFAIK it is not valid to have whitespaces between 'mailTo:' and the email address. Also, usually `mailTo` is written as `mailto`.

test/unit/utils/page.spec.js
```diff
@@ -112,8 +112,34 @@ describe('Page Utils', () => {
...
+    createElement('h1', 'mailTo: john@test.com, also found: another@test.com !', document)
+    const result = getContextElements('h1', 10).length
+    expect(result).to.be.eql(10)
+  })
+
+  it('getContextElements should properly encode the context elements found', function () {
+    createElement('p', 'Some dummy text', document)
+    const size = getContextElements('', 1000).length
+    // Let's simply check if the format is ok
+    expect(size % 4).to.be.eql(0)
```
3link, line 131: We do not ensure this anywhere, we just do `base64UrlEncode(collectedElements).slice(0, contextElementsLength)`.

3link (changes requested):

src/utils/consts.js
```diff
@@ -8,3 +8,4 @@ export const PEOPLE_VERIFIED_LS_ENTRY = '_li_duid'
 export const DEFAULT_IDEX_EXPIRATION_HOURS = 1
 export const DEFAULT_IDEX_AJAX_TIMEOUT = 5000
 export const DEFAULT_IDEX_URL = 'https://idx.liadm.com/idex'
+export const DEFAULT_CONTEXT_ELEMENT_LENGTH = 5000
```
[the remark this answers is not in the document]
paulosimsk: As Gennaro, in my understanding should be the input source for the config, i think it makes sense to remove it.

test/unit/utils/page.spec.js
paulosimsk: Updating as per the previous comment

src/utils/email.js
paulosimsk: It makes sense to broaden the matching capability of the regex, but we should try matching emails anchored but the `mailto:` and also anchored by let's say, empty spaces, right?

src/utils/email.js
```diff
@@ -11,6 +12,7 @@ export function isEmail (s) {
 }
 
 const emailLikeRegex = /"([^"]+(@|%40)[^"]+[.][a-z]*(\s+)?)(\\"|")/
+export const emailLikeNoDoubleQuotesRegex = /([\w.+-]+(@|%40)[\w-]+\.[\w.-]+)/
```
melgenek, line 15: can we unify these regexps into one?

src/utils/page.js
```diff
@@ -38,6 +41,32 @@ export function getPage (win = window) {
...
+    var collectedElements = _collectElementsText(contextSelectors, contextElementsLength, win)
+    return base64UrlEncode(collectedElements)
+  }
+}
+
+function _collectElementsText (contextSelectors, contextElementsLength, win = window) {
+  const collectedElements = win.document.querySelectorAll(contextSelectors)
+  var i = 0
+  var collectedString = ''
+  while (i < collectedElements.length) {
```
melgenek, line 60:
```suggestion
  for (let i =0; i < collectedElements.length; i++) {
```

src/utils/page.js
```diff
@@ -38,6 +41,32 @@ export function getPage (win = window) {
...
+    var nextElement = replaceEmailsWithHashes(collectedElements[i].outerHTML, emailLikeNoDoubleQuotesRegex).stringWithoutRawEmails
+    var n = nextElement.length + collectedString.length
+    if (4 * Math.ceil(n / 3.0) < contextElementsLength) collectedString = collectedString + nextElement
```
3link, line 63: You should compare the amount of bytes in the utf-8 encoded string here, not the length of the string itself.

melgenek: where does this `4/3` proportion come from? Why does it work for bytes?

3link: base64 encodes 3 bytes into 4 bytes representing ASCII characters.

melgenek (approved):

cjs/initializer.js
```diff
@@ -713,6 +746,31 @@ function getPage() {
...
+function getContextElements(contextSelectors, contextElementsLength) {
+  var win = arguments.length > 2 && arguments[2] !== undefined ? arguments[2] : window;
+  if (!contextSelectors || contextSelectors === '' || !contextElementsLength) {
+    return '';
+  } else {
+    var collectedElements = _collectElementsText(contextSelectors, contextElementsLength, win);
+    return base64UrlEncode(collectedElements);
+  }
+}
+function _collectElementsText(contextSelectors, contextElementsLength) {
```
wi101, line 758: do we need to add `win` as an argument?

src/utils/email.js
```diff
@@ -10,7 +11,7 @@ export function isEmail (s) {
   return emailRegex().test(s)
 }
 
-const emailLikeRegex = /"([^"]+(@|%40)[^"]+[.][a-z]*(\s+)?)(\\"|")/
+const emailLikeRegex = /([\w.+-]+(@|%40)[\w-]+\.[\w.-]+)/
```
3link, line 14: Are we sure we are OK with not matching non-ASCII characters?

src/utils/email.js
```diff
@@ -39,3 +40,44 @@ export function listEmailsInString (s) {
...
+ * @param originalString
+ * @returns {{hashesFromOriginalString: HashedEmail[], stringWithoutRawEmails: string}}
+ */
+export function replaceEmailsWithHashes (originalString) {
+  const emailsInString = listEmailsInString(originalString)
+  const hashes = []
+  for (let i = 0; i < emailsInString.length; i++) {
+    const email = emailsInString[i]
+    const emailHashes = hashEmail(email)
+    originalString = originalString.replace(email, emailHashes.md5)
```
3link, line 76: Lets' use another variable here.  Returning `stringWithoutRawEmails: originalString` looks strange at the first sight.

src/utils/page.js
```diff
@@ -38,6 +41,37 @@ export function getPage (win = window) {
   return detectedPageUrl
 }
 
+/**
+ * @return {string|undefined}
+ */
+export function getContextElements (contextSelectors, contextElementsLength, win = window) {
```
3link, line 47: Why is `win` needed here? I saw that we are doing it in other functions functions also and the generated code does some magic like `var win = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : window;` But why?

test/it/live-connect.spec.js
```diff
@@ -243,4 +243,14 @@ describe('LiveConnect', function () {
 
     expect(server.getBakerHistory().length).to.eq(2)
   })
+
+  // - Main page http://bln.test.liveintent.com:3001/self-triggering-page
+  it('should send the collected context elements from page', function () {
```
3link, line 248: 👍

test/unit/enricher/page.spec.js
```diff
@@ -13,12 +13,26 @@ describe('PageEnricher', () => {
...
   })
 
-  it('should return the url and the referrer of the page"', function () {
-    const state = {}
+  it('should return the url, referrer and the contextElements of the page', function () {
+    var newHeadline = document.createElement("h1")
+    var content = document.createTextNode("Some header")
+    newHeadline.appendChild(content)
+    var newP = document.createElement("p")
+    var newContentEmail = document.createTextNode("mailto:john@test.com, also found: another@test.com !")
```
3link, line 21: 👍

3link (approved):

PR #54, opened 2021-11-05
CM-300 - Change the parameter ordering in LiveConnect

3link:
Short description if any.

Author Todo List:

- [ ] Build in CI passes
- [ ] Self-Review

### DIFF
--- a/src/pixel/state.js
+++ b/src/pixel/state.js
@@ -101,7 +101,10 @@ const _pMap = {
   },
   referrer: referrer => {
     return asStringParam('refr', referrer)
-  },
+  }
+}
+
+const _pMapLowPriority = {
   contextElements: contextElements => {
     return asStringParam('c', contextElements)
   }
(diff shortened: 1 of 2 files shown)


gipeshka (approved):

wi101 (approved):

src/pixel/state.js
```diff
@@ -154,11 +157,20 @@ export function StateWrapper (state) {
...
+   * @returns {string [][]}
+   * @private
+   */
+  function _tuples (parameterMap) {
     let array = []
     Object.keys(_state).forEach((key) => {
       const value = _state[key]
-      if (_pMap[key]) {
-        const params = _pMap[key](value)
+      if (parameterMap[key]) {
```
melgenek, line 172: I would just make `_pMap` a list of tuples instead of a map. Then the order is preserved. And implement this method by traversing _pMap instead of the state.

3link: This should work. Let me change that.

src/pixel/state.js
```diff
@@ -52,60 +52,111 @@ import { toParams } from '../utils/url'
...
+  [
+    'liveConnectId',
+    fpc => {
+      return asStringParam('duid', fpc)
+    }
+  ],
+  [
+    'trackerName',
+    tn => {
+      return asStringParam('tna', tn)
```
3link, line 77: We previously had  `asStringParam('tna', tn || 'unknown')` here. From the code it seems, we never returned `unknown` though. Also the tests started failing when `unknown` was returned as default in case the real value is not present. LP defaults to `lp` when `tna` is empty.

3link:
@wi101 @gipeshka @melgenek 1 re-review please

gipeshka (approved):

wi101 (approved):

melgenek (approved):

PR #64, opened 2022-04-08
CM-432: [MVP] -  Adjust LiveConnect To Respect The gdprApplies Flag

wi101:
[CM-432](https://liveintent.atlassian.net/browse/CM-432)

Short description if any.

Author Todo List:

- [ ] Build in CI passes
- [ ] Self-Review


src/model/storage-strategy.js
```diff
@@ -5,7 +5,8 @@
 const StorageStrategy = {
   cookie: 'cookie',
   localStorage: 'ls',
-  none: 'none'
+  none: 'none',
+  disabled: 'disabled'
```
3link, line 9: We should describe it in [CONFIGURATION_OPTIONS](https://github.com/LiveIntent/live-connect/blob/master/CONFIGURATION_OPTIONS.md) and also mention that  it will the the behaviour in case gdprApplies is set to true.

3link (changes requested):

src/handlers/storage-handler.js
```diff
@@ -40,7 +40,7 @@ export function StorageHandler (storageStrategy, externalStorageHandler) {
     }
   }
 
-  const _orElseNoOp = (fName) => strEqualsIgnoreCase(storageStrategy, StorageStrategy.none) ? _noOp : _externalOrError(fName)
+  const _orElseNoOp = (fName) => strEqualsIgnoreCase(storageStrategy, StorageStrategy.none) || strEqualsIgnoreCase(storageStrategy, StorageStrategy.disabled) ? _noOp : _externalOrError(fName)
```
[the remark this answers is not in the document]
wi101: in that case I worry if the old strategy before it's set to disabled is `none`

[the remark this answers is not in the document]
wi101: in that case I would not change `none` --> `disabled` WDYT?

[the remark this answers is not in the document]
wi101: is it correct now?

src/pixel/state.js
```diff
@@ -135,7 +135,12 @@ const _pArray = [
   [
     'gdprApplies',
     gdprApplies => {
-      return asStringParamTransform('gdpr', gdprApplies, (s) => s ? 1 : 0)
+      var params = []
+      params.push(asStringParamTransform('gdpr', gdprApplies, (s) => s ? 1 : 0))
+      params.push(asStringParam('n3pc', gdprApplies))
+      params.push(asStringParam('n3pc_ttl', gdprApplies))
+      params.push(asStringParam('nbakers', gdprApplies))
```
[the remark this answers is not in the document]
wi101: adjusted it

src/idex/identity-resolver-nocache.js
```diff
@@ -26,6 +26,7 @@ function _responseReceived (successCallback) {
 export function IdentityResolver (config, calls) {
   try {
     const nonNullConfig = config || {}
+    const n3pc = nonNullConfig.gdprApplies
```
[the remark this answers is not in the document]
wi101: The second link doesn't redirect me to a specific place, will try it on Monday

src/pixel/state.js
```diff
@@ -185,9 +206,14 @@ export function StateWrapper (state) {
   /**
    * @type {State}
    */
+  let stateWithAdditionalFields = {}
   let _state = {}
   if (state) {
-    _state = _safeFiddle(state)
+    if ('gdprApplies' in state) {
+      const gdprApplies = state.gdprApplies
+      stateWithAdditionalFields = { ...state, n3pc: gdprApplies, n3pc_ttl: gdprApplies, nbakers: gdprApplies }
```
3link, line 214: gdprApplies evaluates to 1 even if the value if not a boolean - e.g. s ? 1 : 0 with s="a" would evaluate to 1. We should compute the values of n3pc, n3pc_ttl and nbakers so that it always matches the value of gdprApplies: gdprApplies == 0 => all three are false, gdprApplies == 1 => all three are true.

wi101: Initially gdprApplies is a boolean and during initializing the state will be converted to 1 or 0 in "gdpr" parameter

src/idex/identity-resolver-nocache.js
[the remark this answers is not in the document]
3link: The second link is actually working, but GitHub considers the change set in that file too big and does not show it. But I meant that: https://github.com/LiveIntent/live-connect/blob/master/src/standard-live-connect.js.

src/handlers/read-storage-handler.js
```diff
@@ -24,7 +24,7 @@ const _noOp = () => undefined
 export function StorageHandler (storageStrategy, externalStorageHandler) {
   const errors = []
   function _externalOrError (functionName) {
-    const hasExternal = externalStorageHandler && externalStorageHandler[functionName] && isFunction(externalStorageHandler[functionName])
```
mschuwalow, line 27: I would move the `!strEqualsIgnoreCase(storageStrategy, StorageStrategy.disabled)` to the beginning of the and chain as it's a bit more explicit that we really don't care about the externalStorageHandler in that case

src/handlers/storage-handler.js
```diff
@@ -31,7 +31,7 @@ export function StorageHandler (storageStrategy, externalStorageHandler) {
   const errors = []
 
   function _externalOrError (functionName) {
-    const hasExternal = externalStorageHandler && externalStorageHandler[functionName] && isFunction(externalStorageHandler[functionName])
+    const hasExternal = externalStorageHandler && externalStorageHandler[functionName] && isFunction(externalStorageHandler[functionName]) && !strEqualsIgnoreCase(storageStrategy, StorageStrategy.disabled)
```
mschuwalow, line 34: same here

src/minimal-live-connect.js
```diff
@@ -26,10 +28,12 @@ import { CallHandler } from './handlers/call-handler'
 function _minimalInitialization (liveConnectConfig, externalStorageHandler, externalCallHandler) {
   try {
     const callHandler = CallHandler(externalCallHandler)
-    const storageHandler = StorageHandler(liveConnectConfig.storageStrategy, externalStorageHandler)
+    const storageStrategy = liveConnectConfig.gdprApplies ? StorageStrategy.disabled : liveConnectConfig.storageStrategy
+    const storageHandler = StorageHandler(storageStrategy, externalStorageHandler)
```
mschuwalow, line 32: We probably need to set the new parameters explicitly here when `gdprApplies` is true

src/standard-live-connect.js
```diff
@@ -130,14 +132,15 @@ function _standardInitialization (liveConnectConfig, externalStorageHandler, ext
     eventBus.init()
     const callHandler = CallHandler(externalCallHandler)
     errorHandler.register(liveConnectConfig, callHandler)
-
-    const storageHandler = StorageHandler(liveConnectConfig.storageStrategy, externalStorageHandler)
+    const storageStrategy = liveConnectConfig.gdprApplies ? StorageStrategy.disabled : liveConnectConfig.storageStrategy
```
mschuwalow, line 135: We probably need to set the new parameters explicitly here when `gdprApplies` is true

CONFIGURATION_OPTIONS.md
```diff
@@ -38,6 +38,14 @@ There's also an option for the module to never create any first party identifier
 ```
 By setting this flag, LiveConnect will never write anything into any storage, but will still attempt to read from it.
 
+There's also an option for the module to never create any first & third party identifiers and this wi, and that can be achieved by setting this parameter as follows:
```
3link, line 41: Looks like part of the sentence got lost.

CONFIGURATION_OPTIONS.md
```diff
@@ -38,6 +38,14 @@ There's also an option for the module to never create any first party identifier
...
+```javascript
+{
+  storageStrategy: 'disabled'
+}
+```
+This flag is internally set when the `config.gdprApplies` is true, LiveConnect will never write anything into any storage, but will still attempt to read from it.
```
3link, line 47: The idea of `disabled` strategy is to also not read anything.

src/enrichers/state-with-no-cookies-config.js
```diff
@@ -0,0 +1,11 @@
+export function enrich (state) {
+  if (state && state.gdprApplies != null) {
+    const gdprApplies = !!state.gdprApplies
+    return {
+      ...state,
```
3link, line 5: We should not build the final state here. Let's implement this enricher the same way others are implemented. The result then can be merged with the current state as we already do it with other enrichers.

3link (changes requested):

src/handlers/read-storage-handler.js
```diff
@@ -24,7 +24,7 @@ const _noOp = () => undefined
...
+    const hasExternal = !strEqualsIgnoreCase(storageStrategy, StorageStrategy.disabled) && externalStorageHandler && externalStorageHandler[functionName] && isFunction(externalStorageHandler[functionName])
```
3link, line 27: We will send an errors in line 31, even though we decided to go with _noOp because of the disabled strategy.

3link (changes requested):

src/handlers/read-storage-handler.js
3link: Also, when sending errors we need to make sure, that we also send the privacy related parameters. Currently, the privacy related flags are appended after the errors callback was registered.

3link (changes requested):

CONFIGURATION_OPTIONS.md
```diff
@@ -38,6 +38,14 @@ There's also an option for the module to never create any first party identifier
 ```
 By setting this flag, LiveConnect will never write anything into any storage, but will still attempt to read from it.
 
+There's also an option for the module to never create any first & third party identifiers, and that can be achieved by setting this parameter as follows:
```
3link, line 41:
```suggestion
There's also an option for the module to never read or create any first & third party identifiers. That can be achieved by setting this parameter as follows:
```

3link (commented):
Please bring back package-lock.json

src/enrichers/no-cookies-config.js
```diff
@@ -0,0 +1,12 @@
+import { isNonEmpty } from '../utils/types'
+
+export function enrich (state) {
+  if (isNonEmpty(state) && isNonEmpty(state.gdprApplies)) {
+    const gdprApplies = !!state.gdprApplies
+    return {
+      n3pc: gdprApplies,
```
3link, line 7: We also should skip computing context elements if gdprApplies.

3link: I propose to simplify the config enrichment part and only compute one field `privacyApplies`. Parameters sent to LivePixel then depend on this parameter and the same parameter can be used to detect if we need to skip computing context elements.

src/handlers/read-storage-handler.js
```diff
@@ -25,11 +25,15 @@ export function StorageHandler (storageStrategy, externalStorageHandler) {
...
-      errors.push(functionName)
+    if (strEqualsIgnoreCase(storageStrategy, StorageStrategy.disabled)) {
       return _noOp
+    } else {
+      if (hasExternal) {
+        return externalStorageHandler[functionName]
+      } else {
+        errors.push(functionName)
+        return _noOp
+      }
```
mschuwalow, line 36:
```suggestion
    } else if (hasExternal) {
        return externalStorageHandler[functionName]
      } else {
        errors.push(functionName)
        return _noOp
```

src/handlers/storage-handler.js
```diff
@@ -32,11 +32,15 @@ export function StorageHandler (storageStrategy, externalStorageHandler) {
...
-      errors.push(functionName)
+    if (strEqualsIgnoreCase(storageStrategy, StorageStrategy.disabled)) {
       return _noOp
+    } else {
+      if (hasExternal) {
+        return externalStorageHandler[functionName]
+      } else {
+        errors.push(functionName)
+        return _noOp
+      }
```
mschuwalow, line 43:
```suggestion
    } else if (hasExternal) {
        return externalStorageHandler[functionName]
      } else {
        errors.push(functionName)
        return _noOp
```

src/enrichers/privacy-config.js
```diff
@@ -0,0 +1,10 @@
+import { isNonEmpty } from '../utils/types'
+
+export function enrich (state) {
+  if (isNonEmpty(state) && isNonEmpty(state.gdprApplies)) {
+    const privacyMode = !!state.gdprApplies
+    return {
+      privacyMode: privacyMode
```
mschuwalow, line 7: It could sense to make privacyMode an enum so it's easier to expand for ccpa support later

mschuwalow (approved):

src/enrichers/privacy-config.js
3link: We could do it. Though for now it is a boolean indicating that we run in privacy mode independently of the actual privacy framework.

PR #65, opened 2022-05-04
CM-446: Update the way additional attributes are serialized for idx endpoint

wi101:
[Ticket link](https://liveintent.atlassian.net/browse/CM-446)
Short description if any.

Author Todo List:

- [x] Build in CI passes
- [x] Self-Review

### DIFF
--- a/src/utils/types.js
+++ b/src/utils/types.js
@@ -114,11 +114,11 @@ export function asStringParamWhen (param, value, predicate) {
 export function mapAsParams (paramsMap) {
   if (paramsMap && isObject(paramsMap)) {
     const array = []
-    Object.entries(paramsMap).forEach(([key, arr]) => {
+    Object.entries(paramsMap).forEach(([key, values]) => {
       const value = paramsMap[key]
       if (value && !isObject(value) && value.length) {
-        if (Array.isArray(arr)) {
-          arr.forEach(id => array.push([encodeURIComponent(key), encodeURIComponent(id)]))
+        if (Array.isArray(values)) {
+          values.forEach(id => array.push([encodeURIComponent(key), encodeURIComponent(id)]))
         } else {
           array.push([encodeURIComponent(key), encodeURIComponent(value)])
         }
--- a/src/utils/types.js
+++ b/src/utils/types.js
@@ -114,11 +114,11 @@ export function asStringParamWhen (param, value, predicate) {
 export function mapAsParams (paramsMap) {
   if (paramsMap && isObject(paramsMap)) {
     const array = []
-    Object.entries(paramsMap).forEach(([key, values]) => {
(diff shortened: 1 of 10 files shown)


mschuwalow (approved):

src/utils/types.js
```diff
@@ -114,9 +114,15 @@ export function asStringParamWhen (param, value, predicate) {
...
+    Object.entries(paramsMap).forEach(([key, values]) => {
       const value = paramsMap[key]
-      value && !isObject(value) && value.length && array.push([encodeURIComponent(key), encodeURIComponent(value)])
+      if (value && !isObject(value) && value.length) {
+        if (Array.isArray(values)) {
+          values.forEach(id => array.push([encodeURIComponent(key), encodeURIComponent(id)]))
+        } else {
+          array.push([encodeURIComponent(key), encodeURIComponent(value)])
+        }
+      }
```
3link, line 125: This is a bit inconsistent: you use a combination of `entries` with `paramsMap[key]`. One of these should be enough. 

Also, using `entries` instead of `keys` increased the size of the generated tag even though the functionality barely changed. I propose to continue using `keys`:

```
export function mapAsParams (paramsMap) {
  if (paramsMap && isObject(paramsMap)) {
    const array = []
    Object.keys(paramsMap).forEach((key) => {
      const value = paramsMap[key]
      if (value && !isObject(value) && value.length) {
        if (isArray(value)) {
          value.forEach(entry => array.push([encodeURIComponent(key), encodeURIComponent(entry)]))
        } else {
          array.push([encodeURIComponent(key), encodeURIComponent(value)])
        }
      }
    })
    return array
  } else {
    return []
  }
}
```

test/unit/idex/identity-resolver-nocache.spec.js
```diff
@@ -72,6 +72,19 @@ describe('IdentityResolver without cache', () => {
...
+    const successCallback = (responseAsJson) => {
+      expect(requestToComplete.url).to.eq('https://idx.liadm.com/idex/unknown/any?duid=987&qf=0.1&resolve=age&resolve=gender')
+      expect(errors).to.be.empty()
+      expect(responseAsJson).to.be.eql(response)
+      done()
+    }
+    identityResolver.resolve(successCallback, () => {}, { qf: '0.1', resolve: ['age', 'gender'] })
+    requestToComplete.respond(200, { 'Content-Type': 'application/json' }, JSON.stringify(response))
+  })
+
```
3link, line 87: 👍

3link (changes requested):

src/utils/types.js
wi101: you're right, it looks better that way, adjusted the code! thank you!

3link (approved):

PR #66, opened 2022-05-05
Add tests for attaching publisher id to identity resolution requests

3link:
Author Todo List:

- [x] Build in CI passes
- [x] Self-Review

### DIFF
--- a/test/unit/idex/identity-resolver-nocache.spec.js
+++ b/test/unit/idex/identity-resolver-nocache.spec.js
@@ -72,6 +72,19 @@ describe('IdentityResolver without cache', () => {
     requestToComplete.respond(200, { 'Content-Type': 'application/json' }, JSON.stringify(response))
   })
 
+  it('should attach publisher id', function (done) {
+    const response = { id: 112233 }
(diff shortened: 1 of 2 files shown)


test/unit/idex/identity-resolver-nocache.spec.js
```diff
@@ -85,6 +85,19 @@ describe('IdentityResolver without cache', () => {
     requestToComplete.respond(200, { 'Content-Type': 'application/json' }, JSON.stringify(response))
   })
 
+  it('should attach publisher id', function (done) {
+    const response = { id: 112233 }
+    const identityResolver = IdentityResolver({ peopleVerifiedId: '987', identityResolutionConfig: { publisherId: '123' } }, calls)
```
wi101, line 90: am not sure if it is the important, but should the publisher Id be a number, should we check it with this format? or it doesn't matter?
```scala
    identityResolutionConfig: { publisherId: 123 } 
```

3link: It does not matter. It is appended to the query part of the URL as is without any assumptions about the type.

3link: Changed it since we generate the config with a number.

wi101 (approved):

PR #71, opened 2022-07-27
RESEARCH - Supporting Multiple LiveConnect Initializations On The Same Page

3link:
**Current State**
There is the LiveConnect module (in the following text the term module will be used) that we distribute and that is used by third parties like Prebid. There is the LiveConnect version that we  distribute via our CDN (in the following text the term builder will be used).

The module accepts a configuration in its constructor functions. The builder expects a configuration either being provided via window.liQ.push or by setting the global property window.LI.

Currently, when LiveConnect (LC) is initialized multiple times on a page, the configuration that was used for the first initialization will prevail, the other ones will be ignored:

- if the module is initialized using config A and then there is an attempt to initialize it with config B, the module will check if there is an already initialized LC and if so, just return it; [it will also check if the config of the initialized LC deviates from the provided config in appId, collector URL or wrapper name](https://github.com/LiveIntent/live-connect/blob/de67aa6b81eaa6254dd3eb95e5b596fe47847df2/src/standard-live-connect.js#L71) and emit an error if so

- if LC is initialized (either via module or builder) and a config is published via window.liQ.push({config: …}), [this config will be ignored and an error message emitted](https://github.com/LiveIntent/live-connect/blob/de67aa6b81eaa6254dd3eb95e5b596fe47847df2/src/standard-live-connect.js#L55)

**Problem**

We have customers that have LC installed on their websites and have a partnership with ad tech companies that also place LC on these websites. This results, depending on the order in which the scripts are loaded, in a LC being initialized with a configuration that belongs to the ad tech company and not to our customer itself. 

The assumption for this work is that the partners do not provide an appId in their configurations. This leads to losing information in the data sent to LivePixel - e.g. not having the appId or not starting the SyncContainer.

**Goal**

When LC is loaded multiple times with different configurations on the same page, the resulting LC must have the configuration that has an appId, if such a configuration was provided on the page.

**Proposal**
In this proposal the following changes have been made:

The module was adjusted as follows:
- current window.liQ is overwritten when the provided config has higher priority than the current one (the presence of appId is the discriminator for now)
- adjusted the module to collect and expose configs a config that have been submitted via window.liQ push when a LC instance was already running

The builder was adjusted as follows:
- the builder uses a local bus for initialization
- the builder evaluates all available configurations upon its initialization and selects one with the highest priority (the one with the highest priority from the ones known to the currently running LC + window.LI + window.liQ - in that order; if no config with appId is present, select the first one or wait for one to appear)
- the builder provides a callback when calling the module; this callback is called whenever the module overwrites the current window.liQ; it publishes events into privacy topics and starts the SyncContainer if the applied config permits it

**Notes**
With this proposal, the global event bus remains untouched when window.liQ is overwritten. That allows to keep all subscribers and does not overcomplicates the code. However, if the old LC is not garbage collected after the new one has been initialized, we will see one error reported to LivePixel twice. After thinking more about it, it is not a big deal. We can, if we see that it is an issue, identify and remove "internal" topic listeners if required when we overwrite window.liQ.

Also, the SyncContainer may be executed multiple times if there are multiple configs with appId and enabled sync (and no gdpr). This will cause no harm. In case the first initialized LC does not have an appId or has the sync disabled or has gdpr enabled, but the second one has an appId, enabled sync and no gdpr, the sync container will work based on the config of the first LC. In either case, if the sync would be executed according to one of the used configs, it will be executed. 

**TODOs**: 
- [ ] define scenarios with combinations of old/new module and builder, SyncContainer, configurations as well as the times these configurations become available and execute thorough manual tests
- [ ] thorough automated testing

Related PRs:
- https://github.com/LiveIntent/liveconnectbuilder/pull/61
- https://github.com/LiveIntent/synccontainer/pull/31

Author Todo List:

- [ ] Build in CI passes
- [ ] Self-Review

### DIFF
--- a/src/config/config-manager.js
+++ b/src/config/config-manager.js
@@ -3,10 +3,6 @@ export default function ConfigManager () {
   this.seenConfigs = []
 }
 
-export function hasPriorityOver (configA, configB) {
-  return qualifiedConfig(configA) && !qualifiedConfig(configB)
-}
-
 export function qualifiedConfig (config) {
   return !!config.appId
 }
--- a/src/config/config-manager.js
+++ b/src/config/config-manager.js
@@ -1,0 +1,22 @@
+
+export default function ConfigManager () {
+  this.seenConfigs = []
+}
+
+export function hasPriorityOver (configA, configB) {
+  return qualifiedConfig(configA) && !qualifiedConfig(configB)
+}
+
+export function qualifiedConfig (config) {
+  return !!config.appId
+}
+
+ConfigManager.prototype = {
+  push: function (config) {
+    this.seenConfigs.push(config)
+  },
+
+  configs: function () {
+    return this.seenConfigs
+  }
+}
--- a/src/standard-live-connect.js
+++ b/src/standard-live-connect.js
@@ -130,9 +130,10 @@ function _getInitializedLiveConnect (liveConnectConfig) {
  * @returns {StandardLiveConnect}
  * @private
  */
-function _standardInitialization (liveConnectConfig, externalStorageHandler, externalCallHandler) {
+function _standardInitialization (liveConnectConfig, externalStorageHandler, externalCallHandler, localBus) {
   try {
     eventBus.init()
+    _emitShutdownMessage(liveConnectConfig)
     const callHandler = CallHandler(externalCallHandler)
     const configWithPrivacy = merge(liveConnectConfig, privacyConfig(liveConnectConfig))
     errorHandler.register(configWithPrivacy, callHandler)
@@ -150,7 +151,12 @@ function _standardInitialization (liveConnectConfig, externalStorageHandler, ext
     console.log('LiveConnect.postManagedState', postManagedState)
     const syncContainerData = merge(configWithPrivacy, { peopleVerifiedId: postManagedState.data.peopleVerifiedId })
     const onPixelLoad = () => emitter.send(C.PIXEL_SENT_PREFIX, syncContainerData)
-    const onPixelPreload = () => emitter.send(C.PRELOAD_PIXEL, '0')
+    const onPixelPreload = () => {
+      if (localBus) {
+        localBus.emit(C.PRELOAD_PIXEL, '0')
+      }
+      emitter.send(C.PRELOAD_PIXEL, '0')
+    }
     const pixelClient = new PixelSender(configWithPrivacy, callHandler, onPixelLoad, onPixelPreload)
     const resolver = IdentityResolver(postManagedState.data, storageHandler, callHandler)
     const _configManager = _initializeConfigManager(liveConnectConfig)
@@ -172,6 +178,10 @@ function _standardInitialization (liveConnectConfig, externalStorageHandler, ext
   }
 }
 
+function _emitShutdownMessage (config) {
+  emitter.send(C.LC_SHUTDOWN, config)
+}
+
 function _initializeConfigManager (config) {
   if (window.liQ && window.liQ.configManager) {
     const configManager = window.liQ.configManager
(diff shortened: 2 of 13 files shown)


src/standard-live-connect.js
```diff
@@ -103,17 +104,19 @@ function _processArgs (args, pixelClient, enrichedState) {
...
+function _getInitializedLiveConnect (liveConnectConfig, initializationCallback) {
   try {
     if (window && window.liQ && window.liQ.ready) {
-      const mismatchedConfig = window.liQ.config && _configMatcher(window.liQ.config, liveConnectConfig)
-      if (mismatchedConfig) {
-        const error = new Error()
-        error.name = 'ConfigSent'
-        error.message = 'Additional configuration received'
-        emitter.error('LCDuplication', JSON.stringify(mismatchedConfig), error)
+      if (qualifiedConfig(window.liQ.config) && !initializationCallback) {
```
wi101, line 110: Does this mean we will emit an error like we used to but only for the clients that use live-connect module
and this won't be the same for the builder?

mschuwalow:
Unless I'm blind, this doesn't solve the problem of reconfiguring / creating a new liveconnect instance with the new configuration.

src/standard-live-connect.js
```diff
@@ -41,18 +41,19 @@ import { resolve as decisionsResolve } from './manager/decisions'
...
+import ConfigManager, { qualifiedConfig } from './config/config-manager'
 
 const hemStore = {}
-function _pushSingleEvent (event, pixelClient, enrichedState) {
+function _pushSingleEvent (event, pixelClient, enrichedState, configManager) {
   if (!event || !isObject(event)) {
     emitter.error('EventNotAnObject', 'Received event was not an object', new Error(event))
   } else if (event.config) {
-    emitter.error('StrayConfig', 'Received a config after LC has already been initialised', new Error(event))
+    configManager.push(event.config)
```
wi101, line 56: for those who use npm module, in order to allow multiple configuration, here in this part should we pick the best configuration and send the event to pixelClient?

3link: The NPM module is not meant to be configured via window.liQ.push. Instead, the config must be passed explicitly when calling the create function. The code here collects the configs that arrive via window.liQ.push because it may be required by the LC initialized in the next step. Specifically, it covers this case:

- init NPM module (the config is provided explicitly)
- window.liQ.push({config: {...}})
- load builder script

src/standard-live-connect.js
3link: I adjusted the condition a little. The idea to prioritize the config with an appId that comes from the builder higher than the current config because the builder will also initialize the privacy properties and the SyncContainer.

src/config/config-manager.js
```diff
@@ -0,0 +1,18 @@
+
+export default function ConfigManager () {
+  this.seenConfigs = []
+}
+
+export function qualifiedConfig (config) {
```
mschuwalow, line 6: Let's assign a numeric value to the 'goodness' of particular configuration. Than we can just compare which one is better on startup

src/standard-live-connect.js
```diff
@@ -103,17 +104,20 @@ function _processArgs (args, pixelClient, enrichedState) {
...
+function _getInitializedLiveConnect (liveConnectConfig, initializationCallback) {
   try {
     if (window && window.liQ && window.liQ.ready) {
-      const mismatchedConfig = window.liQ.config && _configMatcher(window.liQ.config, liveConnectConfig)
-      if (mismatchedConfig) {
-        const error = new Error()
-        error.name = 'ConfigSent'
-        error.message = 'Additional configuration received'
-        emitter.error('LCDuplication', JSON.stringify(mismatchedConfig), error)
+      if ((!qualifiedConfig(liveConnectConfig)) ||
```
mschuwalow, line 110: We could also go with another option. When we detect we have 2 instances we can shut down the previous instance, unregistering all listeners.
We also can create a new topic which we publish to whenever we are starting a new instance

3link: > We could also go with another option. When we detect we have 2 instances we can shut down the previous instance, unregistering all listeners.

We will lose external listeners that way. To be able to keep them, we would need to be able to distinguish them from the internal ones. Which is doable, but maybe not worth the overhead.

mschuwalow: It should not be too difficult. We would keep using the same global bus, just have a topic to publish cancellation commands to. The external consumer would not notice anything

3link:
Prioritization of configuration is not the way to go. This is the approach we agreed upon: https://github.com/LiveIntent/live-connect/pull/74

PR #72, opened 2022-08-18
[CM-565] Use latest code from liveconnectbuilder for tests

mschuwalow:
Short description if any.

Author Todo List:

- [x] Build in CI passes
- [x] Self-Review


test/unit/manager/decisions.spec.js
```diff
@@ -42,8 +42,8 @@ describe('DecisionsManager for stored decisions', () => {
 
   it('should return unique decision ids', function () {
     const decisionId = uuid()
-    storage.setCookie('lidids.123', decisionId, { domain: 'something.example.com' })
```
mschuwalow, line 45: Looks like this one is a bug / typo from very long ago

PR #75, opened 2022-09-08
[No ticket] Don't keep generated code in repository

mschuwalow:
Short description if any.

Author Todo List:

- [x] Build in CI passes
- [x] Self-Review


3link (commented):
Can you please check that we do not package the removed cjs and esm variants? I did not look into it in detail, but the two seem not to be results of intermediate steps but just two variant using different approaches for modules. If we do not need them, we could also skip generating them.

mschuwalow:
We are not packaging them as far as I can tell. I wanted to keep them around to not break something subtle in our packaging steps.

But you are right, we can probably remove it entirely. updated

3link (approved):

peixunzhang (approved):

.npmignore
```diff
@@ -0,0 +1,9 @@
+*.DS_Store
```
mschuwalow, line 1: If this file is present, it should be used to filter files during packaging instead of the .gitignore

PR #76, opened 2022-09-08
[CM-586] Do not resolve uid2 when running in privacy mode

mschuwalow:
https://liveintent.atlassian.net/browse/CM-586

Author Todo List:

- [x] Build in CI passes
- [x] Self-Review


3link (approved):

3link:
Like the refactoring. 👍

leonelcuevas (approved):

src/idex/common.js
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    return {
+      resolve: (successCallback, errorCallback, additionalParams) => {
+        try {
+          unsafeResolve(successCallback, errorCallback, additionalParams)
+        } catch (e) {
+          console.error('IdentityResolve', e)
+          errorCallback()
+          fromError('IdentityResolve', e)
```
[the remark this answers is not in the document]
mschuwalow: I guess so? :D Kept them the same as they were.

The string passed to `fromError` is what shows up here https://prod-grafana.liveintent.com/d/3WtwayWWz/livepixel?orgId=2&refresh=10s&viewPanel=13, I don't know whether the value has any meaning beyond it

mschuwalow:
Talked with Dejan and we are going to make a change in idres that will make the nonId/unifiedId handling cleaner. Will keep this unmerged until then

src/idex/common.js
```diff
@@ -0,0 +1,178 @@
...
+/**
+ * @type {Cache}
+ */
+export const noopCache = {
+  get: (key) => {
+    return null
+  },
+  set: (key, value) => {
+
+  }
```
wi101, line 61: I would do it this way, but as you prefer
```suggestion
  get: (key) => null,
  set: (key, value) => undefined
```

src/idex/common.js
```diff
@@ -0,0 +1,178 @@
...
+      tuples.push(asStringParam(retrievedIdentifier.name, retrievedIdentifier.value))
+    })
+
+    const attributeResolutionAllowed = (attribute) => {
+      if (attribute === 'uid2') {
+        return !config.privacyMode
+      } else {
+        return true
+      }
+    }
```
wi101, line 102:
```suggestion
    const attributeResolutionAllowed = (attribute) => {
      attribute === 'uid2' ? !config.privacyMode : true
    }
```

wi101 (approved):
LGTM, just tiny suggestions

3link (approved):

PR #81, opened 2022-09-26
[CM-586] Consider expires header when caching idres responses

mschuwalow:
https://liveintent.atlassian.net/browse/CM-587

Short description if any.

Author Todo List:

- [x] Build in CI passes
- [x] Self-Review

### DIFF
--- a/src/idex/common.js
+++ b/src/idex/common.js
@@ -13,7 +13,7 @@ import { DEFAULT_IDEX_AJAX_TIMEOUT, DEFAULT_IDEX_URL, DEFAULT_REQUESTED_ATTRIBUT
 /**
  * @return {Cache}
  */
-export function storageHandlerBackedCache (expirationHours, domain, storageHandler) {
+export function storageHandlerBackedCache (defaultExpirationHours, domain, storageHandler) {
   const IDEX_STORAGE_KEY = '__li_idex_cache'
 
   function _cacheKey (rawKey) {
@@ -34,12 +34,12 @@ export function storageHandlerBackedCache (expirationHours, domain, storageHandl
         return cachedValue
       }
     },
-    set: (key, value) => {
+    set: (key, value, expiresAt) => {
       try {
         storageHandler.set(
           _cacheKey(key),
           JSON.stringify(value),
-          expiresInHours(expirationHours),
+          expiresAt || expiresInHours(defaultExpirationHours),
           domain
         )
       } catch (ex) {
@@ -54,7 +54,7 @@ export function storageHandlerBackedCache (expirationHours, domain, storageHandl
  */
 export const noopCache = {
   get: (key) => null,
-  set: (key, value) => undefined
+  set: (key, value, expiresAt) => undefined
 }
 
 /**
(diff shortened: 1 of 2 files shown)


3link (approved):

3link:
Rather related to the implementation of the feature in general: There are now two ways of defining what needs to be resolved in LC:
- via the `requestedAttributes` config parameter, which results in `resolve` parameters being send to IdRes: https://github.com/LiveIntent/live-connect/pull/76/files#diff-c6b1818a2e0b7118e720980ab2d8fc3af8270c3e72d85837820f769633a29383R100
- via the `additionalParams` parameter of the `resolve` function which also results in `resolve` parameters being send to IdRes (additionalParams can be something like `{resolve: ['nonId', 'md5', 'uid2']}):` https://github.com/LiveIntent/live-connect/pull/76/files#diff-c6b1818a2e0b7118e720980ab2d8fc3af8270c3e72d85837820f769633a29383R133

Let's get rid of the `requestedAttributes` config parameter. Prebid's user id module could then translate the `requestedAttributesOverrides` config into `additionalParams` for the resolve function.

mschuwalow:
My understanding is that additionalParams is a bit of an escape hatch. We are already exposing other configuration parameters that are more clear / specific for other parameters -- I don't really see a reason to depart from this pattern for this parameter in particular

src/idex/common.js
```diff
@@ -163,3 +165,12 @@ export function makeIdentityResolver (config, calls, cache) {
     }
   }
 }
+
+function responseExpires (response) {
+  if (response && typeof response.getResponseHeader === 'function') {
```
wi101, line 170: You can use `isFunction` from `utils/types.js`

src/idex/common.js
```diff
@@ -108,25 +108,27 @@ export function makeIdentityResolver (config, calls, cache) {
...
           } catch (ex) {
             console.error('Error parsing response', ex)
             fromError('IdentityResolverParser', ex)
           }
         }
-        cache.set(additionalParams, responseObj)
+
+        const expiresAt = responseExpires(response)
+
+        cache.set(additionalParams, responseObj, expiresAt)
```
wi101, line 124: there will be no issue if the expiresAt is undefined?

src/idex/common.js
```diff
@@ -108,25 +108,27 @@ export function makeIdentityResolver (config, calls, cache) {
...
-            responseObj = JSON.parse(response)
+            responseObj = JSON.parse(responseText)
           } catch (ex) {
             console.error('Error parsing response', ex)
             fromError('IdentityResolverParser', ex)
           }
         }
-        cache.set(additionalParams, responseObj)
+
+        const expiresAt = responseExpires(response)
```
wi101, line 122: should it be?
```suggestion
        const expiresAt = responseExpires(responseObj)
```

mschuwalow: We need to extract it from the headers. ResponseText is only the body

src/idex/common.js
mschuwalow: the cache handles it properly
`expiresAt || expiresInHours(defaultExpirationHours)`

wi101 (approved):

PR #84, opened 2022-09-29
[No ticket][TypeScript] Update publishing setup

mschuwalow:
Updates our packaging setup to have a more modern / best practices setup.
After this is merged we will stop emitting the cjs + esm (which is a backwards compat breaking change / we will have one anyway unless we want to generate a fake 'src' folder from typescript).

Instead the following files are created:
```
dist
├── index.d.ts // types that will be picked up by typescript / javascript when importing our library
├── index.js // library compiled to commonjs
└── index.mjs // thin ecmascript wrapper on top of our library
```

Some remarks:

The index.js file is based on ./src/index.ts and allows us to have a very curated import experience for our users.
```
export * from './initializer'
export * from './minimal-live-connect'
export * from './standard-live-connect'

export * as emitter from './utils/emitter'
export * as eventBus from './events/bus'
export * as types from './utils/types'
```

The generated `./dist/index.mjs` file is just a thin wrapper on top of commonjs modules and follows recommended best practices https://nodejs.org/api/packages.html#packages_dual_commonjs_es_module_packages

```
import cjs from './index.js';

export const LiveConnect = cjs.LiveConnect;
export const MinimalLiveConnect = cjs.MinimalLiveConnect;
export const StandardLiveConnect = cjs.StandardLiveConnect;
export const emitter = cjs.emitter;
export const error = cjs.error;
export const eventBus = cjs.eventBus;
export const fromError = cjs.fromError;
export const send = cjs.send;
export const types = cjs.types;
```

The changes to packages.json allow downstream dependencies to automatically pick the correct module for the module system they use:
https://nodejs.org/api/packages.html#conditional-exports
```
  "main": "./dist/index.js",
  "module": "./dist/index.mjs",
  "types": "./dist/index.d.ts",
  "exports": {
    ".": {
      "require": "./dist/index.js",
      "import": "./dist/index.mjs"
    },
    "./package.json": "./package.json"
  },
```

As a consequence of this we no longer need the ugly explicit import in liveconnectbuilder / prebid (i.e. `import { LiveConnect } from 'live-connect-js/esm/initializer.js';`). Instead we can use the nice import syntax we also use for other dependencies:
`import { LiveConnect, MinimalLiveConnect } from 'live-connect-js';` 

Author Todo List:

- [x] Build in CI passes
- [x] Self-Review


.npmignore
```diff
@@ -1,9 +0,0 @@
-*.DS_Store
```
mschuwalow, line 1: Replaced by `files` in package.json

babel.config.js
```diff
@@ -1,12 +1,7 @@
 module.exports = {
   presets: [
     [
-      '@babel/preset-env',
-      {
-        // useBuiltIns: 'usage',
```
mschuwalow, line 6: These were unused

leonelcuevas (approved):

PR #85, opened 2022-10-11
Make LiveConnect’s Global Variable Name Configurable

leonelcuevas:
[CM-607](https://liveintent.atlassian.net/browse/CM-607)

Todo List:
- [x] Build in CI passes
- [x] Self-Review


src/enrichers/identifiers-nohash.js
```diff
@@ -5,18 +5,17 @@
...
 import { safeToString, isArray, trim } from '../utils/types'
-import * as emitter from '../utils/emitter'
 
 /**
  * @param {State} state
  * @param {StorageHandler} storageHandler
  * @returns {{hashesFromIdentifiers: HashedEmail[], retrievedIdentifiers: RetrievedIdentifier[]} | {}}
  */
-export function enrich (state, storageHandler) {
+export function enrich (state, storageHandler, messageBus) {
```
3link, line 14: Let's extend the doc with the new parameter.

src/events/event-bus.js
```diff
@@ -0,0 +1,56 @@
...
+
+export function LocalEventBus (size) {
+  return initBus(size)
+}
+
+export function GlobalEventBus (name, size, errorCallback) {
+  try {
+    console.log('events.bus.init')
+    if (!window) {
+      errorCallback(new Error('Bus can only be attached to the window, which is not present'))
```
3link, line 38:
```suggestion
      errorCallback(new Error('Global bus can only be attached to the window, which is not present'))
```

src/enrichers/identifiers.js
```diff
@@ -5,18 +5,17 @@
...
 import { safeToString, isString, isArray } from '../utils/types'
-import * as emitter from '../utils/emitter'
 
 /**
  * @param {State} state
  * @param {StorageHandler} storageHandler
  * @returns {{hashesFromIdentifiers: HashedEmail[], retrievedIdentifiers: RetrievedIdentifier[]} | {}}
  */
-export function enrich (state, storageHandler) {
+export function enrich (state, storageHandler, messageBus) {
```
3link, line 14: We should also extends the doc and add the new parameter.

src/enrichers/people-verified.js
```diff
@@ -1,12 +1,11 @@
...
 import { PEOPLE_VERIFIED_LS_ENTRY } from '../utils/consts'
 
-export function enrich (state, storageHandler) {
+export function enrich (state, storageHandler, messageBus) {
   console.log('people-verified.enrich', state)
   try {
     return { peopleVerifiedId: state.peopleVerifiedId || storageHandler.getDataFromLocalStorage(PEOPLE_VERIFIED_LS_ENTRY) }
   } catch (e) {
-    error.error('PeopleVerifiedEnrich', e.message, e)
+    messageBus.emitError('PeopleVerifiedEnrich', e.message, e)
```
3link, line 8: We actually use `encodeEmitError` here. We then would not need to call `e.message` here, the function would do it for us.

src/events/event-bus.js
```diff
@@ -0,0 +1,56 @@
...
+
+  bus.emitError = function (name, message, e = {}) {
+    const wrappedError = wrapError(name, message, e)
+    bus.emit(C.ERRORS_PREFIX, wrappedError)
+  }
+
+  bus.encodeEmitError = function (name, exception) {
+    bus.emitError(name, exception.message, exception)
+  }
+}
```
3link, line 28: Maybe we can rename the two to have less cryptic names. For example `emitError` to `emitErrorWithMessage` and `encodeEmitError` to `emitErrorExtractMessage`.

wi101: +1 for  `emitError ` --> `emitErrorWithMessage` 
how about leaving `emitError` for `encodeEmitError ` ?

wi101: I will leave it as it is for the old replayEmitter in case customers are emitting errors that way

src/events/event-bus.js
```diff
@@ -0,0 +1,56 @@
...
+
+function extendBusIfNeeded (bus) {
+  if (isNewEmitter(bus)) {
+    return
+  }
+
+  bus.emitError = function (name, message, e = {}) {
+    const wrappedError = wrapError(name, message, e)
+    bus.emit(C.ERRORS_PREFIX, wrappedError)
```
wi101, line 22: I will replace this with the impl of `bus.emitError(name, message, e)`

src/events/event-bus.js
wi101: done

.circleci/config.yml
```diff
@@ -177,10 +177,12 @@ workflows:
           context: live-connect
           type: approval
           requires:
-            - browserstack
+            - simple_test
```
wi101, line 180: this should be rolled back before the PR get merged

CONFIGURATION_OPTIONS.md
```diff
@@ -5,11 +5,25 @@ This field does not have the default value. If it is not set, the appId calculat
...
+```
+#### `distributorId` [Optional]
+This parameter is an identifier for distributors.
+`distributorId` does not have a default value and must be set in case of distributors scripts.
+Note: The distributor script has `globalVarName` property that enables to:
+ * initialize LiveConnect configuration for distributors including the `distributorId`.
+ * send distributors events.
+
+If `distributorId` is provided, the `appId` configuration parameter must not be set.
+Example:
```
3link, line 19: `globalVarName` is independent of `distributorId`. It just allows you to define an arbitrary name for the variable we attach to `window` and that can be used to interact with LiveConnect.

wi101: I adjust it and removed that, 
but for this:
https://github.com/LiveIntent/live-connect/pull/85#discussion_r999023686
 where should we define the `globalVarName` ?

src/standard-live-connect.js
```diff
@@ -105,15 +104,20 @@ function _processArgs (args, pixelClient, enrichedState) {
...
-        emitter.error('LCDuplication', JSON.stringify(mismatchedConfig), error)
+        const eventBus = getAndAttachGlobalBus(liveConnectConfig.globalVarName)
+        eventBus.emitErrorWithMessage('LCDuplication', JSON.stringify(mismatchedConfig), error)
       }
-      return window.liQ
+      return window[liveConnectConfig.globalVarName]
+    }
+    if (liveConnectConfig.globalVarName !== 'liQ') {
+      window.liQ_instances = window.liQ_instances || []
+      window.liQ_instances.push(liveConnectConfig.globalVarName)
```
3link, line 120: Why not adding `liQ` to the array too?

3link: Could we put this functionality somewhere else? It seems to add to the complexity of `_getInitializedLiveConnect`, but does not have anything to do with actually getting an initialized LC.

3link: We could even save references to the initializes LCs instead of their global names.

README.md
```diff
@@ -133,8 +133,8 @@ The `enrichers` folder contains code responsible for extracting specific informa
 ### Identifiers enrichment
 `enrichers/identifiers.js` is responsible for reading the `identifiersToResolve` configuration parameter to read any additional identifiers that customers want to share with us.
 
-## Messaging between components via EventBus (`__li__evt_bus`)
-LiveConnect exposes an object on the window level (`window.__li_evt_bus`) which is responsible for communicating various information based on different fields of interests.
+## Messaging between components via EventBus
+LiveConnect exposes an object on the instance level (`window[globalVarName].eventBus`, where `globalVarName` is the value found in the configuration or `liQ` if none is provided) which is responsible for communicating various information based on different fields of interests.
```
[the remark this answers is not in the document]
3link: I do not see it there :). I also would add it as a regular configuration option independent from the `distributorId`.

CONFIGURATION_OPTIONS.md
3link: It is an ordinary configuration parameter. Just put the description of it into CONFIGURATION_OPTIONS.md.

src/standard-live-connect.js
```diff
@@ -124,69 +124,78 @@ function _getInitializedLiveConnect (liveConnectConfig) {
...
+    window && (window[configuration.globalVarName] = _getInitializedLiveConnect(configuration) || _standardInitialization(configuration, externalStorageHandler, externalCallHandler, eventBus) || queue)
     if (isArray(queue)) {
       for (let i = 0; i < queue.length; i++) {
-        window.liQ.push(queue[i])
+        window[configuration.globalVarName].push(queue[i])
       }
     }
+    window.liQ_instances = window.liQ_instances || []
+
+    if (window.liQ_instances.filter(i => i.config.globalVarName === configuration.globalVarName).length === 0) {
```
3link, line 193: Can we use `filter` here?

src/standard-live-connect.js
3link: Looks good. Left a comment.

src/standard-live-connect.js
leonelcuevas: Do you mean `find`? or `filter` where?

3link: Sorry, I meant `find`. @wi101 told me that it does not work in every browser.

README.md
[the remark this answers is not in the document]
leonelcuevas: It's now added ^^

test-config/wdio.local.chrome.conf.js
```diff
@@ -99,7 +99,7 @@ exports.config = {
   connectionRetryCount: 3,
 
   services: [
-      ['selenium-standalone', { drivers: { chrome: true } }]
+      ['selenium-standalone', { drivers: { chrome: '106.0.5249.61' } }]
```
3link, line 102: Just curious, why a specific version, not the latest?

test/unit/enricher/identifiers.spec.js
```diff
@@ -188,8 +189,9 @@ describe('IdentifiersEnricher', () => {
     const getCookieStub = sandbox.stub(storage, 'getCookie').throws()
     storage.setCookie(COOKIE_NAME, SIMPLE_COOKIE1)
     const state = { identifiersToResolve: [COOKIE_NAME] }
+    const eventBus = LocalEventBus()
 
-    const resolutionResult = identifiersEnricher.enrich(state, storage)
+    const resolutionResult = identifiersEnricher.enrich(state, storage, eventBus)
```
3link, line 194: Why are we mixing busses in this case? There is a local one for the enricher and a global one for the external storage.

src/events/event-bus.js
```diff
@@ -0,0 +1,65 @@
...
+    console.error('events.bus.init', e)
+    errorCallback(e)
+  }
+}
+
+export function getAndAttachGlobalBus (name) {
+  const eventBus = window[name].eventBus || window[C.EVENT_BUS_NAMESPACE]
+  extendBusIfNeeded(eventBus)
+  window[name].eventBus = eventBus
+  return eventBus
```
3link, line 64: This function has the side effect of modifying `window[name]`. Can we split it into a function that determines the bus to use and a direct assignment to `window[name]` on the caller's side?

test/unit/idex/identity-resolver.spec.js
```diff
@@ -2,27 +2,27 @@ import jsdom from 'mocha-jsdom'
...
 
 use(dirtyChai)
 
 describe('IdentityResolver', () => {
   let requestToComplete = null
+  const messageBus = LocalEventBus()
   let errors = []
   let callCount = 0
-  const storage = StorageHandler('cookie', externalStorage)
+  const storage = StorageHandler('cookie', ExternalStorage)
```
3link, line 18: Can we use the same bus for all components that are tested together by a test?

test/unit/manager/identifiers.spec.js
```diff
@@ -1,14 +1,16 @@
...
 import jsdom from 'mocha-jsdom'
 import dirtyChai from 'dirty-chai'
 import { StorageHandler } from '../../../src/handlers/storage-handler'
+import { LocalEventBus } from '../../../src/events/event-bus'
 
 use(dirtyChai)
 
-const storage = StorageHandler('cookie', externalStorage)
+const messageBus = LocalEventBus()
+const storage = StorageHandler('cookie', ExternalStorage, messageBus)
```
3link, line 13: This is a mix again.

src/events/event-bus.js
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+function initBus (size) {
+  if (!size) {
+    size = 5
+  }
+  return new E(size)
+}
+
+function extendBusIfNeeded (bus) {
```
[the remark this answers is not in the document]
leonelcuevas: There is a unit test for that already ([here](https://github.com/LiveIntent/live-connect/blob/cm-607/test/unit/events/bus.spec.js#L40)), but through the GlobalEventBus function. Is that what you meant or do you mean to test something else?

test-config/wdio.local.chrome.conf.js
leonelcuevas: In our case we apparently didn't have the latest version so we couldn't test locally, that's why we had to use that specific version. I will change it back and add a comment in case someone finds the same problem.

test/unit/enricher/identifiers.spec.js
leonelcuevas: Yes, I wasn't aware of this mixup, I tried to fix it for all cases, hopefully I didn't miss any

src/events/event-bus.js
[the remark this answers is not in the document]
3link: I should have been more specific. I actually meant the case where we have an old LC initialized and attempt to initialize a new one. In that case, the bus should be wrapped and `eventBus` field should be attached to `window.globalVarName`.

src/events/event-bus.js
```diff
@@ -0,0 +1,64 @@
...
+  const isNewEmitter = isFunction(bus.emitErrorWithMessage) && isFunction(bus.emitError)
```
mschuwalow, line 24: imo, I would just inline the condition

src/minimal-live-connect.js
```diff
@@ -52,15 +54,23 @@ function _minimalInitialization (liveConnectConfig, externalStorageHandler, exte
...
  */
-export function MinimalLiveConnect (liveConnectConfig, externalStorageHandler, externalCallHandler) {
+export function MinimalLiveConnect (liveConnectConfig, externalStorageHandler, externalCallHandler, externalEventBus) {
   console.log('Initializing LiveConnect')
   try {
-    window && (window.liQ = window.liQ || [])
     const configuration = (isObject(liveConnectConfig) && liveConnectConfig) || {}
-    return _minimalInitialization(configuration, externalStorageHandler, externalCallHandler)
+    configuration.globalVarName = configuration.globalVarName || 'liQ'
+    window && (window[configuration.globalVarName] = window[configuration.globalVarName] || [])
```
mschuwalow, line 66: can we do this with an if? I really dislike this assignment inside a boolean expression

src/standard-live-connect.js
mschuwalow: find should work given that we are using babel. It would be very surprising if it does

mschuwalow (approved):

CONFIGURATION_OPTIONS.md
```diff
@@ -5,11 +5,21 @@ This field does not have the default value. If it is not set, the appId calculat
...
 ```javascript
 {
-  appId:"a-00xx"
+  appId: "a-00xx"
+}
+```
+#### `distributorId` [Optional]
+This parameter is an identifier for distributors.
+It does not have a default value and must be set in case of distributors scripts.
+If `distributorId` is provided, the `appId` configuration parameter must not be set.
```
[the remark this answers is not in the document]
mschuwalow: +1 let's add a check for this

3link (approved):

wi101 (approved):

PR #94, opened 2023-01-05
Investigate and fix decode urls in Live-connect

wi101:
## [CM-630](https://liveintent.atlassian.net/browse/CM-630)
The logs: [here](https://search-streams-production-ggatbwema5ghxbi5aqpcvkcile.us-east-1.es.amazonaws.com/_plugin/kibana/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-4h,to:now))&_a=(columns:!(_source),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:afb18c50-a95e-11ea-b7d9-89ab4e3c16e1,key:details.trackerVersion,negate:!f,params:(query:v2.5.1),type:phrase),query:(match_phrase:(details.trackerVersion:v2.5.1))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:afb18c50-a95e-11ea-b7d9-89ab4e3c16e1,key:errorDetails.name,negate:!f,params:(query:DecisionsResolve),type:phrase),query:(match_phrase:(errorDetails.name:DecisionsResolve)))),index:afb18c50-a95e-11ea-b7d9-89ab4e3c16e1,interval:auto,query:(language:kuery,query:''),sort:!()))
shows errors: Error while managing decision ids: URIError: URI malformed

And when you look to the `requestDetails.originUrl` they contain `%` even they aren't encoded. And in this case the url doesn't need to be decoded.

Example:

open this page: 

[https://www.bluenile.com/hk/zh/diamonds?gclid=CjwKCAiAh9qdBhAOEiwAvxIok639kgT6vvuWV9-EkJL8w9K70leaL5ARCmhvkWxBerVNkpI3nOBiaxoC8soQAvD_BwE&click_id=384939136&utm_source=google&utm_medium=text&utm_campaign=Google_%7C_GC_%7C_HK_%7C_Traditional_Chinese_%7C_Text_%7C_Non-Brand_%7C_ENG_%7C_NA_%7C_Engagement_%7&utm_content=Diamonds%3A_Price&utm_term=%E4%B8%80+%E5%85%8B%E6%8B%89+%E9%91%BD%E7%9F%B3+%E5%83%B9%E6%A0%BC](https://www.bluenile.com/hk/zh/diamonds?gclid=CjwKCAiAh9qdBhAOEiwAvxIok639kgT6vvuWV9-EkJL8w9K70leaL5ARCmhvkWxBerVNkpI3nOBiaxoC8soQAvD_BwE&click_id=384939136&utm_source=google&utm_medium=text&utm_campaign=Google_%257C_GC_%257C_HK_%257C_Traditional_Chinese_%257C_Text_%257C_Non-Brand_%257C_ENG_%257C_NA_%257C_Engagement_%257&utm_content=Diamonds%253A_Price&utm_term=%25E4%25B8%2580+%25E5%2585%258B%25E6%258B%2589+%25E9%2591%25BD%25E7%259F%25B3+%25E5%2583%25B9%25E6%25A0%25BC)

We get an error:

```
Error while managing decision ids
URIError: URI malformed
    at decodeURIComponent (<anonymous>)
    at he (https://b-code.liadm.com/a-0389.min.js:1:9458...
```

The reason is when the decision calls urlParams it tries to decode the url because it contains `%` on its values even it's not properly encoded: https://github.com/LiveIntent/live-connect/blob/master/src/utils/url.js#L15

Author Todo List:

- [x] Build in CI passes
- [x] Self-Review

[CM-630]: https://liveintent.atlassian.net/browse/CM-630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/utils/url.js
+++ b/src/utils/url.js
@@ -31,7 +31,16 @@ function _convert (v) {
   return _isBoolean(_isNull(_isNum(v)))
 }
 
-function _urlParams (url) {
+function _parseParam (params, key) {
+  if (isArray(params[key])) {
+    params[key] = params[key].map(v => _convert(_decode(`${v}`)))
+  } else {
+    params[key] = _convert(_decode(`${params[key]}`))
+  }
+  return params[key]
+}
+
+function _allParams (url) {
   let questionMarkIndex, queryParams, historyIndex
   const obj = {}
   if (!url || (questionMarkIndex = url.indexOf('?')) === -1 || !(queryParams = url.slice(questionMarkIndex + 1))) {
@@ -52,22 +61,12 @@ function _urlParams (url) {
 }
 
 export function urlParams (url) {
-  const params = _urlParams(url)
-  Object.keys(params).forEach((k) => {
-    if (isArray(params[k])) {
-      params[k] = params[k].map(v => _convert(_decode(`${v}`)))
-    } else {
-      params[k] = _convert(_decode(`${params[k]}`))
-    }
-  })
+  const params = _allParams(url)
+  Object.keys(params).forEach((k) => _parseParam(params, k))
   return params
 }
 
 export function getQueryParameter (url, name) {
-  const params = _urlParams(url)
-  if (isArray(params[name])) {
-    return params[name].map(v => _convert(_decode(`${v}`)))
-  } else {
-    return _convert(_decode(`${params[name]}`))
-  }
+  const params = _allParams(url)
+  return _parseParam(params, name)
 }
--- a/src/utils/url.js
+++ b/src/utils/url.js
@@ -12,7 +12,7 @@ export const toParams = (tuples) => {
 }
 
 function _decode (s) {
-  return s.indexOf('%') === -1 ? s : decodeURIComponent(s)
+  return s.indexOf('%') === -1 ? s : decodeValue(s)
 }
 
 function _isNum (v) {
@@ -58,6 +58,10 @@ function _allParams (url) {
     }
   })
   return obj
+}
+
+export function decodeValue (v) {
+  return v.replace(/(%[\dA-F]{2})+/gi, decodeURIComponent)
 }
 
 export function urlParams (url) {
--- a/src/utils/url.js
+++ b/src/utils/url.js
@@ -12,7 +12,14 @@ export const toParams = (tuples) => {
 }
 
 function _decode (s) {
-  return s.indexOf('%') === -1 ? s : decodeURIComponent(s)
+  if (s.indexOf('%') === -1) return s
+  else {
+    try {
+      decodeURIComponent(s)
+    } catch (e) {
+      return s
+    }
+  }
 }
 
 function _isNum (v) {
--- a/src/utils/url.js
+++ b/src/utils/url.js
@@ -11,10 +11,6 @@ export const toParams = (tuples) => {
   return acc
 }
 
-function _decode (s) {
-  return s.indexOf('%') === -1 ? s : decodeValue(s)
-}
-
 function _isNum (v) {
   return isNaN(+v) ? v : +v
 }
@@ -34,9 +30,9 @@ function _convert (v) {
 function _parseParam (params, key) {
   if (params[key]) {
     if (isArray(params[key])) {
-      params[key] = params[key].map(v => _convert(_decode(v)))
+      params[key] = params[key].map(v => _convert(decodeValue(v)))
     } else {
-      params[key] = _convert(_decode(params[key]))
+      params[key] = _convert(decodeValue(params[key]))
     }
     return params[key]
   }
--- a/src/utils/url.js
+++ b/src/utils/url.js
@@ -1,4 +1,4 @@
-import { isArray, isFunction } from './types'
+import { isArray } from './types'
 
 export const toParams = (tuples) => {
   let acc = ''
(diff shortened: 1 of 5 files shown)


3link (changes requested):

3link:
There is a theoretical case where we will use a malformed decision id with this approach. However, I am not sure, if this case can actually appear in reality since dids usually have a format of a UUID (encoded string = decoded string).

Maybe it is possible to turn it around and instead of failing when decoding parameters that are not used anyway, the code could attempt to specifically decode the `li_did`, ignore the others and fail if it is not possible.

src/utils/url.js
```diff
@@ -42,11 +48,15 @@ export function urlParams (url) {
   }
   queryParams.split('&').forEach(function (query) {
     if (query) {
-      query = ((query = query.split('=')) && query.length === 2 ? query : [query[0], 'true']).map(_decode)
+      query = ((query = query.split('=')) && query.length === 2 ? _queryNameValue(query, predicate) : [query[0], 'true'])
```
3link, line 51: This will only decode the parameter when the predicate matches and return the original value otherwise. 

I actually meant that we can create a function that allows to extract only one particular parameter (`getQueryParameter(url, name)`). And the code would also attempt to only decode this one parameter, do not process other parameters at all and fail if it is not possible. The function can then be is used [here](https://github.com/LiveIntent/live-connect/blob/master/src/manager/decisions.js#L31-L32) to extract lidids.

src/utils/url.js
```diff
@@ -31,7 +31,13 @@ function _convert (v) {
   return _isBoolean(_isNull(_isNum(v)))
 }
 
-export function urlParams (url) {
+function _queryNameValue (query, predicate) {
+  const name = _decode(query[0])
+  const value = (isFunction(predicate) && predicate.length === 1 && predicate(name)) ? _decode(query[1]) : query[1]
```
mschuwalow, line 36: I would not do the `predicate.length === 1` check or at least change it to be `predicate.length >= 1` as it's fine to pass a function that expects more arguments in js world

src/utils/url.js
mschuwalow: I agree that reading a single parameter is probably the way to go.

An alternative would be to go the same way as our cookie library and manually only decode the urlencoded parts.
`value.replace(/(%[\dA-F]{2})+/gi, decodeURIComponent)`
That is less brittle than our approach.

source: https://github.com/js-cookie/js-cookie/blob/main/src/converter.mjs#L7

wi101: when I try it with this value 
```
'https://www.bluenile.com/hk/zh/diamonds?gclid=CjwKCAiAh9q&click_id=12&utm_source=google&utm_campaign=Google_%7C_GC_%7C_HK_%7C_Traditional_Chinese_%7C_Text_%7C_Non-Brand_%7C_ENG_%7C_NA_%7C_Engagement_%7&utm_content=Diamonds%3A_Price&utm_term=%E4%B8%80+%E5%85%8B%E6%'
```
it will return the same error:
```
Uncaught URIError: URI malformed
```

wi101: Good suggestion though
I tried the `getQueryParameter` solution

src/utils/url.js
wi101: makes sense, we won't use the predicate, decided to go with getQueryParameter suggestion

src/utils/url.js
```diff
@@ -42,11 +51,26 @@ export function urlParams (url) {
...
+export function urlParams (url) {
+  const params = _allParams(url)
+  Object.keys(params).forEach((k) => _parseParam(params, k))
+  return params
+}
+
+export function getQueryParameter (url, name) {
+  const params = _allParams(url)
+  return _parseParam(params, name)
+}
```
3link, line 76: The treatment of the 'undefined' string seems to be odd. For example 
```
 getQueryParameter('https://www.bluenile.com/hk/zh/diamonds?name=undefined', 'name')
```

will give you `null` as value of `name` instead of 'undefined'.

wi101: it was because of the converter there https://github.com/LiveIntent/live-connect/blob/master/src/utils/url.js#L23

3link: Yes, but shouldn't we get the value instead? It is a corner case, but still seems weird.

wi101: we could change the converter to return undefined in that case to forward the exact same value

wi101: done

3link: Hmm, it looks like ti was a desired behavior: https://github.com/LiveIntent/live-connect/blob/master/test/unit/utils/url.spec.js#L29. Revert?

wi101: reverted and now it will return undefined only if the parameter is not defined

src/utils/url.js
```diff
@@ -31,7 +31,18 @@ function _convert (v) {
   return _isBoolean(_isNull(_isNum(v)))
 }
 
-export function urlParams (url) {
+function _parseParam (params, key) {
+  if (params[key]) {
+    if (isArray(params[key])) {
+      params[key] = params[key].map(v => _convert(_decode(`${v}`)))
+    } else {
+      params[key] = _convert(_decode(`${params[key]}`))
```
3link, line 39: You now should be able to get rid of the string building part here.

wi101: done

3link (approved):
With minor suggestions.

src/utils/url.js
```diff
@@ -12,7 +12,7 @@ export const toParams = (tuples) => {
 }
 
 function _decode (s) {
```
mschuwalow, line 14: There is not need for _decode anymore, you can just use decodeValue

mschuwalow (approved):

PR #101, opened 2023-01-18
[CM-667] Cleanup & Reenable lint step

mschuwalow:
Short description if any.

Author Todo List:

- [x] Build in CI passes
- [x] Self-Review


src/handlers/call-handler.ts
```diff
@@ -9,7 +9,7 @@ export function CallHandler (externalCallHandler: ExternalCallHandler, eventBus:
   function _externalOrError (functionName: string) {
     const hasExternal = externalCallHandler && externalCallHandler[functionName] && isFunction(externalCallHandler[functionName])
     if (hasExternal) {
-      return externalCallHandler[functionName]
+      return externalCallHandler[functionName].bind(externalCallHandler)
```
wi101, line 12: 🤔  why do we need to return the externalCallHandler if the functionName isn't defined?

src/windows.d.ts
```diff
@@ -1,7 +1,12 @@
 import { ILiveConnect } from './types'
 
 declare global {
+  type XDomainRequest = any
+
   interface Window {
+    // eslint-disable-next-line camelcase
     liQ_instances?: ILiveConnect[]
+    liQ?: ILiveConnect
+    XDomainRequest?: { new (): XDomainRequest; prototype: XDomainRequest; create(): XDomainRequest; };
```
wi101, line 10: was it required to define `XDomainRequest` ?
it sounds interesting because in [liveconnectbuilder](https://github.com/LiveIntent/liveconnectbuilder/blob/master/src/utils/calls.js#L33) the `calls.js` is using it and I couldn't find from where it will be initialized

src/handlers/call-handler.ts
wi101: I see thanks for the explanation!

src/standard-live-connect.ts
```diff
@@ -116,7 +116,7 @@ function _standardInitialization (liveConnectConfig: LiveConnectConfig, external
   }
 }
 
-export function StandardLiveConnect (liveConnectConfig: LiveConnectConfig, externalStorageHandler: ExternalStorageHandler, externalCallHandler: ExternalCallHandler, externalEventBus: EventBus): ILiveConnect {
+export function StandardLiveConnect (liveConnectConfig: LiveConnectConfig, externalStorageHandler: ExternalStorageHandler, externalCallHandler: ExternalCallHandler, externalEventBus?: EventBus): ILiveConnect {
```
wi101, line 119: we can do that at the initializer and the minimalLC

src/types.ts
```diff
@@ -78,7 +78,7 @@ export type IdentityResultionResult = object
 export interface ExternalCallHandler {
     ajaxGet?: (
         url: string,
-        onSuccess: (responseText: string) => void,
+        onSuccess: (responseText: string, response: any) => void,
```
wi101, line 81:
```suggestion
        onSuccess: (responseText: string, response?: any) => void,
```

wi101 (approved):
minor suggestions otherwise LGTM thanks!

src/types.ts
mschuwalow: we actually always have it

src/standard-live-connect.ts
mschuwalow: MinimalLiveConnect should already have it, but seems like I missed it for LiveConnect.

By the time we are calling the initialisers we always have an eventbus, so no need there

wi101: you're right!

src/types.ts
wi101: oh ok because I saw that previously it wasn't there

mschuwalow: yeah, it was missing in the types

PR #103, opened 2023-01-25
[No Ticket] Cleanup pass and enable strict mode

mschuwalow:
Short description if any.

Author Todo List:

- [x] Build in CI passes
- [x] Self-Review


test/unit/idex/identity-resolver.spec.ts
```diff
@@ -37,15 +38,15 @@ describe('IdentityResolver', () => {
...
+    const identityResolver = IdentityResolver.make({}, storageHandler, calls, eventBus)
     const successCallback = (responseAsJson) => {
       expect(callCount).to.be.eql(1)
       expect(errors).to.be.empty()
       expect(responseAsJson).to.be.eql(response)
       expect(requestToComplete.url).to.eq('https://idx.liadm.com/idex/unknown/any')
       expect(responseAsJson).to.be.eql(response)
       expect(callCount).to.be.eql(1)
-      expect(storageHandler.getCookie('__li_idex_cache')).to.be.eq(JSON.stringify(response))
+      expect(storageHandler.getCookie('__li_idex_cache_e30')).to.be.eq(JSON.stringify(response))
```
mschuwalow, line 49: The cache key for no provided additional parameters changed. (As we are using `{}` instead of `undefined`)

I don't think this is a problem as we will just re-resolve, but I can revert this change if desired

src/pixel/state.ts
```diff
@@ -157,19 +73,21 @@ export class Query {
 
 export class StateWrapper {
   data: State
-  eventBus: EventBus
+  eventBus?: EventBus
 
-  constructor (state: State, eventBus: EventBus) {
+  constructor (state: State, eventBus?: EventBus) {
```
leonelcuevas, line 78: Don't we always provide a bus? Why does it need to be optional?

src/enrichers/people-verified.ts
```diff
@@ -1,9 +1,10 @@
 import { PEOPLE_VERIFIED_LS_ENTRY } from '../utils/consts'
-import { EventBus, IMinimalStorageHandler, State } from '../types'
+import { EventBus, State } from '../types'
+import { MinimalStorageHandler } from '../handlers/storage-handler'
 
-export function enrich (state: State, storageHandler: IMinimalStorageHandler, eventBus: EventBus) {
+export function enrich (state: State, storageHandler: MinimalStorageHandler, eventBus: EventBus): State {
   try {
-    return { peopleVerifiedId: state.peopleVerifiedId || storageHandler.getDataFromLocalStorage(PEOPLE_VERIFIED_LS_ENTRY) }
+    return { peopleVerifiedId: state.peopleVerifiedId || storageHandler.getDataFromLocalStorage(PEOPLE_VERIFIED_LS_ENTRY) || undefined }
```
3link, line 7: Why is explicit `undefined` required here? - Uniformly returning undefined instead of a mix null/unified?

src/initializer.ts
```diff
@@ -6,10 +6,10 @@ import { EventBus, ExternalCallHandler, ExternalStorageHandler, ILiveConnect, Li
 import { EVENT_BUS_NAMESPACE } from './utils/consts'
 import { GlobalEventBus } from './events/event-bus'
 
-export function LiveConnect (liveConnectConfig: LiveConnectConfig, externalStorageHandler: ExternalStorageHandler, externalCallHandler: ExternalCallHandler, mode: string, externalEventBus?: EventBus): ILiveConnect {
+export function LiveConnect (liveConnectConfig: LiveConnectConfig, externalStorageHandler: ExternalStorageHandler, externalCallHandler: ExternalCallHandler, mode: string, externalEventBus?: EventBus): ILiveConnect | null {
```
3link, line 9: Why is the bus optional?

test/unit/idex/identity-resolver.spec.ts
3link: I think it is fine.

3link:
Have to do a second pass. The PR is too large for one pass.

src/enrichers/people-verified.ts
mschuwalow: yep, exactly

src/initializer.ts
mschuwalow: We are creating our own when it's not passed in https://github.com/LiveIntent/live-connect/pull/103/files/bfcffb59ee3111e0f9d47c6e4b564b0b3ceed0ad#diff-16aa336504940f7a9c705777cc8e89acd054c3e54f132f6af961ada25aa6165dL11

src/enrichers/identifiers.ts
```diff
@@ -1,18 +1,21 @@
 import { replaceEmailsWithHashes } from '../utils/email'
 import { safeToString, isString, isArray } from '../utils/types'
-import { EventBus, HashedEmail, IMinimalStorageHandler, State } from '../types'
+import { EventBus, HashedEmail, State, RetrievedIdentifier } from '../types'
+import { MinimalStorageHandler } from '../handlers/storage-handler'
 
-export function enrich (state: State, storageHandler: IMinimalStorageHandler, eventBus: EventBus) {
+export function enrich (state: State, storageHandler: MinimalStorageHandler, eventBus?: EventBus): State {
```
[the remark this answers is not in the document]
mschuwalow: We can. I personally like no having to pass a bus if I don't care about getting the errors emitted there, but I can also revert this and adjust the tests.

Making the bus mandatory would have the advantage that it's impossible to 'forget' it. Which way do you prefer?

src/pixel/state.ts
mschuwalow: Mostly missing in a lot of the tests. Same thinking as here: https://github.com/LiveIntent/live-connect/pull/103/files/bfcffb59ee3111e0f9d47c6e4b564b0b3ceed0ad#r1087556144

Let's decide whether we want these mandatory or not, will adjust accordingly

src/enrichers/identifiers.ts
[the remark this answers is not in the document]
3link: I prefer to have a bus because of the reason you mention.

src/enrichers/people-verified.ts
3link: OK!

src/initializer.ts
3link: Ah yes.

src/enrichers/identifiers.ts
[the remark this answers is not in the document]
leonelcuevas: +1, I would also prefer to have it mandatory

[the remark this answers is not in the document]
mschuwalow: ok, updated 👍

src/events/error-pixel.ts
```diff
@@ -39,11 +38,11 @@ export function asErrorDetails (e: any): State {
...
     return {
       errorDetails: {
-        message: _truncate(e.message),
-        name: _truncate(e.name),
+        message: _truncate(e.message) || '',
+        name: _truncate(e.name) || '',
         stackTrace: _truncate(e.stack),
         lineNumber: _asInt(e.lineNumber),
-        lineColumn: _asInt(e.lineColumn),
+        columnNumber: _asInt(e.columnNumber),
```
wi101, line 45: good catch

src/events/event-bus.ts
```diff
@@ -1,13 +1,15 @@
+// @ts-nocheck
```
wi101, line 1: why?

src/handlers/call-handler.ts
```diff
@@ -1,28 +1,46 @@
-import { isFunction } from '../utils/types'
-import { EventBus, ExternalCallHandler, ICallHandler } from '../types'
+import { EventBus, ExternalCallHandler } from '../types'
+import { WrappingContext } from '../utils/wrapping'
 
-const _noOp = () => undefined
+interface WrappedExternalCallHandler {
+  ajaxGet: (
+    url: string,
+    onSuccess: (responseText: string, response: unknown) => void,
```
wi101, line 7:
```suggestion
    onSuccess: (responseText: string, response ?: unknown) => void,
``````

File: src/handlers/storage-handler.ts [hunk 9]
```diff
@@ -1,76 +1,132 @@
...
-    if (strEqualsIgnoreCase(storageStrategy, StorageStrategy.disabled)) {
-      return _noOp
-    } else if (hasExternal) {
-      return externalStorageHandler[functionName].bind(externalStorageHandler)
-    } else {
-      errors.push(functionName)
-      return _noOp
+const noop = () => undefined
+
+function wrapRead<T extends object, K extends keyof T & string> (wrapper: WrappingContext<T>, storageStrategy: StorageStrategy, functionName: K) {
```
wi101, line 21: I would rename it to something else as it is called also in the wrapWriter
should we call it`wrap` ?

src/types.ts
```diff
@@ -1,16 +1,9 @@
...
-}
-
-export interface ErrorDetails {
-    message: string,
-    name: string,
+export interface ErrorDetails extends Error {
     stackTrace?: string,
     lineNumber?: number,
-    lineColumn?: number,
+    columnNumber?: number
```
wi101, line 6:
```suggestion
    columnNumber?: number,
```

src/handlers/storage-handler.ts
mschuwalow: Write is generally stronger than read, so IMO it feels fine to implement write permission checks in terms of read. Other opinions?

src/handlers/call-handler.ts
mschuwalow: no need, the callback will always be provided with the argument

src/events/event-bus.ts
mschuwalow: The file is not properly typed yet. We can do this in a followup (it's a bit tricky though because of all the window manipulation in there)

src/handlers/call-handler.ts
wi101: but when it is called it uses only the first param
https://github.com/LiveIntent/live-connect/blob/master/src/pixel/sender.ts#L57

src/handlers/storage-handler.ts
wi101: You're right +1

wi101 (approved):

leonelcuevas (approved):

src/handlers/call-handler.ts
mschuwalow: Using only one is fine. The questionmark would signal that the function is only provided 1 argument sometimes, even if it asks for 2.

3link (approved):

PR #109, opened 2023-02-22
CM-732. Make the way the page URL is collected configurable

leonelcuevas:
## [CM-732](https://liveintent.atlassian.net/browse/CM-732)

In order to make the way the page URL is collected configurable, the following changes are needed:

1. A new configuration that allows to pick between the URL scraping mode, with 2 modes: 
    - full - collect the full URL (default), 
    - no_page - do not collect the page
2. A second configuration where users can provide a regular expression for filtering out matching query parameters (no filtering by default)

For reference see the [spec doc](https://liveintent.atlassian.net/wiki/spaces/PL/pages/3543072807/Url+Scraping+in+LiveConnect).

Author Todo List:

- [x] Build in CI passes
- [x] Self-Review

[CM-732]: https://liveintent.atlassian.net/browse/CM-732?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/utils/url.ts
```diff
@@ -87,7 +87,45 @@ export function urlParams(url: string): Record<string, ParsedParam | ParsedParam
...
+  hash: string
+  host: string
+  hostname: string
+  pathname: string
+  port: string
+  protocol: string
+  search: string
+
+  constructor (url: string) {
+    // eslint-disable-next-line
```
leonelcuevas, line 112: I was getting an error about not needing to escape the `/`, but having 2 in a row makes it a comment 😅

package-lock.json
```diff
@@ -1,8 +1,19796 @@
 {
   "name": "live-connect-js",
   "version": "4.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
mschuwalow, line 4: Our ci runs npm npm v6.14.16, which should work fine with the new version

src/model/url-collection-mode.ts
```diff
@@ -0,0 +1,6 @@
+export type UrlCollectionMode = 'full' | 'no_page'
+
+export const UrlCollectionModes: Record<string, UrlCollectionMode> = {
```
mschuwalow, line 3: Does autocompletion work for the fields of this? If not, we can just make it a plain record

src/utils/url.ts
mschuwalow: it should be a comment 👍

src/pixel/url-collector.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+  const url = new ParsedUrl(urlString)
+  return paramsToDelete(url, conf)
+}
+
+export function isPageRemoved(urlString: string | undefined, conf: LiveConnectConfig): boolean {
+  if (!urlString || conf.urlCollectionMode !== UrlCollectionModes.noPageUrl) {
+    return false
+  }
+  const url = new ParsedUrl(urlString)
```
mschuwalow, line 26: Why do we not trust our configuration?

src/pixel/url-collector.ts
```diff
@@ -0,0 +1,51 @@
+import { UrlCollectionModes } from '../model/url-collection-mode'
+import { LiveConnectConfig } from '../types'
+import { ParsedUrl, urlParamsArray } from '../utils/url'
+
+export function collectUrl(urlString: string, conf: LiveConnectConfig): string {
```
mschuwalow, line 5: just pass the whole state here, no need for the url to go as a separate parameter imo

src/pixel/state.ts
```diff
@@ -50,7 +51,12 @@ const paramExtractors: ((state: State) => [string, string][])[] = [
   ifDefined('privacyMode', privacyMode => asStringParamWhen('nb', privacyMode ? 1 : 0, v => v === 1)),
   ifDefined('gdprConsent', gdprConsentString => asStringParam('gdpr_consent', gdprConsentString)),
   ifDefined('referrer', referrer => asStringParam('refr', referrer)),
-  ifDefined('contextElements', contextElements => asStringParam('c', contextElements))
+  ifDefined('contextElements', contextElements => asStringParam('c', contextElements)),
+  state => asStringParamWhen('pu_rp', isPageRemoved(state.pageUrl, state) ? 1 : 0, v => v === 1),
```
mschuwalow, line 55: I would combine both of these and the url one so we are not doing duplicate work

src/pixel/url-collector.ts
```diff
@@ -0,0 +1,51 @@
...
+function paramsToDelete(url: ParsedUrl, conf: LiveConnectConfig): string[] {
+  if (conf.queryParametersFilter === undefined || conf.queryParametersFilter === '') {
+    return []
+  }
+  const urlQueryParameters = urlParamsArray(url.search)
+  const filterRegExp = new RegExp(conf.queryParametersFilter)
+  return urlQueryParameters.map(keyValuePair => keyValuePair[0]).filter(param => filterRegExp.test(param))
+}
+
+function paramsToKeep(url: ParsedUrl, conf: LiveConnectConfig): string {
```
mschuwalow, line 39: Can we define this based on paramsToDelete / the other way around?

test/unit/pixel/state.spec.ts
```diff
@@ -41,7 +42,7 @@ describe('EventComposition', () => {
       eventSource: { eventName: 'viewContent' },
       liveConnectId: '213245',
       trackerName: 'test tracker',
-      pageUrl: 'https://wwww.example.com?sss',
+      pageUrl: 'https://wwww.example.com/?sss',
```
mschuwalow, line 45: Why the escape here?

src/model/url-collection-mode.ts
leonelcuevas: I think it does, didn't think about it, but I believe used it a couple times

src/pixel/url-collector.ts
leonelcuevas: True, I wrote this part of the code first and then noticed they come from the same place, but forgot to change it. Thanks!

src/pixel/state.ts
leonelcuevas: Thanks, I just realized I can do that. At first I thought it wouldn't work properly. Will change it

test/unit/pixel/state.spec.ts
leonelcuevas: The URL class I was using always set the pathname to `/` when empty, but I might be able to revert it now. I will check

src/pixel/url-collector.ts
leonelcuevas: In the state model the url is optional so it can be undefined.

Regarding the `url.pathname.length > 1 ` we want to have a field that tracks when the path is actually removed from the url, not just when the "mode" is `no_page` and since the initial `/` is considered part of the path we need to check if the length is `> 1`.

Just as reference, I named the things in the same way as the js [URL class](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL), which I was using initially, but then I found out it wasn't supported on IE.

src/pixel/state.ts
```diff
@@ -24,7 +25,7 @@ const paramExtractors: ((state: State) => [string, string][])[] = [
   ifDefined('eventSource', source => asParamOrEmpty('se', source, (s) => base64UrlEncode(JSON.stringify(s, replacer)))),
   ifDefined('liveConnectId', fpc => asStringParam('duid', fpc)),
   ifDefined('trackerName', tn => asStringParam('tna', tn)),
-  ifDefined('pageUrl', purl => asStringParam('pu', purl)),
+  state => ifDefined('pageUrl', purl => asStringParam('pu', collectUrl(purl, state)))(state),
```
3link, line 28: Would it be possible to do the URL parsing only if it is required. This way we would not make the script slower for customers that do not use that feature.

src/pixel/state.ts
```diff
@@ -50,7 +51,12 @@ const paramExtractors: ((state: State) => [string, string][])[] = [
   ifDefined('privacyMode', privacyMode => asStringParamWhen('nb', privacyMode ? 1 : 0, v => v === 1)),
   ifDefined('gdprConsent', gdprConsentString => asStringParam('gdpr_consent', gdprConsentString)),
   ifDefined('referrer', referrer => asStringParam('refr', referrer)),
-  ifDefined('contextElements', contextElements => asStringParam('c', contextElements))
+  ifDefined('contextElements', contextElements => asStringParam('c', contextElements)),
+  state => asStringParamWhen('pu_rp', isPageRemoved(state.pageUrl, state) ? 1 : 0, v => v === 1),
+  state => {
+    const blockedParams = blockedQueryParams(state.pageUrl, state)
```
3link, line 57: There are currently multiple places at which the URL is parsed and query parameters are filtered. Would it be possible to do only one single pass for all of these steps: if the config requires adjusting the original URL, parse it and do one pass to return the adjusted URL with the info what has been adjusted - e.g. page removed flag, removed query parameters list.

3link (changes requested):

src/pixel/url-collector.ts
mschuwalow: that makes sense, thanks 👍

3link (approved):

mschuwalow (approved):

wi101 (commented):
Consider to add the new params in https://github.com/LiveIntent/live-connect/blob/master/COLLECTOR_PARAMS.md

PR #111, opened 2023-03-02
CM-714 - Do not attach LC to window if no globalVarName is provided

3link:
https://liveintent.atlassian.net/browse/CM-714

This PR
- creates LiveConnect instances with a local event bus by default instead of the global bus
- does not attach the created LiveConnect instance to the `window` object if no globalVarName is provided instead of attaching `liQ` by default

Related PRs (both will be updated once this PR is merged):
- https://github.com/LiveIntent/Prebid.js/pull/11
- https://github.com/LiveIntent/liveconnectbuilder/pull/93

Author Todo List:

- [x] Adjust documentation (liQ is not automatically used as a fallback globalVarName when none is provided)
- [x] Build in CI passes
- [x] Self-Review

This is the follow-up to get rid of the global state handling in LC: https://liveintent.atlassian.net/browse/CM-756

### DIFF
--- a/CONFIGURATION_OPTIONS.md
+++ b/CONFIGURATION_OPTIONS.md
@@ -223,9 +223,9 @@ Example:
 ```
 
 ##### `globalVarName` [Optional]
+**Usage of this parameter is discouraged and it will be deprecated in a future version of this library.** This parameter allows to define the name for the LiveConnect instance that is attached to the window object. If none is provided, Liveconnect will not be attached to the window object. 
 
-This parameter allows to define the name for the Live Connect instance that is attached to the window object, by default: `liQ`.
-For example `globalVarName: "liQ2"` will make the instance name `window.liQ2` instead of `window.liQ`.
+For example `globalVarName: "liQ2"` will make the instance name `window.liQ2`. 
 
 Example:
 ```javascript
--- a/CONFIGURATION_OPTIONS.md
+++ b/CONFIGURATION_OPTIONS.md
@@ -223,7 +223,7 @@ Example:
 ```
 
 ##### `globalVarName` [Optional]
-**Usage of this parameter is discouraged and it will be deprecated in a future version of this library.** This parameter allows to define the name for the LiveConnect instance that is attached to the window object. If none is provided, Liveconnect will not be attached to the window object. 
+**Usage of this parameter is discouraged and it will be deprecated in a future version of this module.** This parameter allows to define the name for the LiveConnect instance that is attached to the window object. If none is provided, Liveconnect will not be attached to the window object. 
 
 For example `globalVarName: "liQ2"` will make the instance name `window.liQ2`. 
 
--- a/src/minimal-live-connect.ts
+++ b/src/minimal-live-connect.ts
@@ -39,41 +39,34 @@ function _minimalInitialization(liveConnectConfig: LiveConnectConfig, externalSt
 }
 
 function _initializeWithoutGlobalName(liveConnectConfig: LiveConnectConfig, externalStorageHandler: ReadOnlyStorageHandler, externalCallHandler: CallHandler, eventBus: EventBus) {
-  return _minimalInitialization(liveConnectConfig, externalStorageHandler, externalCallHandler, eventBus, (event: unknown) => {})
+  const lc = _minimalInitialization(liveConnectConfig, externalStorageHandler, externalCallHandler, eventBus, (event: unknown) => {})
+  window.liQ_instances = window.liQ_instances || []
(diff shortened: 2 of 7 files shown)


src/minimal-live-connect.ts
```diff
@@ -38,22 +38,35 @@ function _minimalInitialization(liveConnectConfig: LiveConnectConfig, externalSt
...
-    }
-    return lc
+    lc = configuration.globalVarName ?
+      _initializeWithGlobalName(configuration, externalStorageHandler, externalCallHandler, eventBus) :
+      _initializeWithoutGlobalName(configuration, externalStorageHandler, externalCallHandler, eventBus)
   } catch (x) {
     console.error(x)
   }
-  return {}
+  return lc
```
mschuwalow, line 71: do we want to keep returning `{}` here, not undefined in case of errors.
Otherwise we will break code like `if (lc.push) ???`

src/standard-live-connect.ts
```diff
@@ -119,28 +119,44 @@ function _standardInitialization (liveConnectConfig: LiveConnectConfig, external
...
+      lc.push(queue[i])
+    }
+  }
+  
+  window[lc.config.globalVarName] = lc
+
+  window.liQ_instances = window.liQ_instances || []
+  if (window.liQ_instances.filter(i => i.config.globalVarName === lc.config.globalVarName).length === 0) {
+    window.liQ_instances.push(lc)
+  }
```
mschuwalow, line 144: Let's factor this out as it's repeated 4 timers

CONFIGURATION_OPTIONS.md
```diff
@@ -223,9 +223,9 @@ Example:
 ```
 
 ##### `globalVarName` [Optional]
+**Usage of this parameter is discouraged and it will be deprecated in a future version of this module.** This parameter allows to define the name for the LiveConnect instance that is attached to the window object. If none is provided, Liveconnect will not be attached to the window object.
```
mschuwalow, line 226: alternatively, let's keep using liQ as the default for now but allow passing null explicitly to disable attaching.

(though of course I'm in favour of nuking it asap :D)

mschuwalow (approved):

src/standard-live-connect.ts
3link: In one of the intermediate versions, I had it factored out. Decided to do it that way on purpose: in the next step all logic that initializes LC with globalVarName will move into the builder project. Thus I can simple remove the _initializeWithGlobalName functions.

src/minimal-live-connect.ts
3link: Sure. Done in c444ac10d7d88a0521f2af47e50d62383a3706a4

CONFIGURATION_OPTIONS.md
3link: Let's nuke it.

test/unit/minimal-live-connect.spec.ts
```diff
@@ -46,47 +46,78 @@ describe('MinimalLiveConnect', () => {
...
+  it('should expose liQ and liQ_instances', function () {
     expect(window.liQ).to.be.undefined()
-    MinimalLiveConnect({}, storage, calls)
+    expect(window.liQ_instances).to.be.undefined()
+    MinimalLiveConnect({ globalVarName: 'liQ' }, storage, calls)
     expect(window.liQ).to.not.be.undefined()
+    expect(window.liQ_instances).to.not.be.undefined()
+  })
+
+  it('should only add liQ_intances to the window object', function () {
```
wi101, line 57:
```suggestion
  it('should only add liQ_instances to the window object', function () {
```

test/unit/minimal-live-connect.spec.ts
```diff
@@ -46,47 +46,78 @@ describe('MinimalLiveConnect', () => {
...
   it('should expose liQ via the initializer', function () {
     expect(window.liQ).to.be.undefined()
-    LiveConnect({}, storage, calls, 'minimal')
+    LiveConnect({ globalVarName: 'liQ' }, storage, calls, 'minimal')
     expect(window.liQ).to.not.be.undefined()
-    expect(window[EVENT_BUS_NAMESPACE]).to.not.be.undefined()
   })
 
-  it('should accept a single event and put it in the queue', function () {
+  it('should only add liQ_intances to the window object via the initializer', function () {
```
wi101, line 71:
```suggestion
  it('should only add liQ_instances to the window object via the initializer', function () {
```

wi101 (approved):

leonelcuevas (approved):

PR #113, opened 2023-03-13
[No Ticket] add global bus

mschuwalow:
Short description if any.

Author Todo List:

- [x] Build in CI passes
- [x] Self-Review

### DIFF
--- a/src/standard-live-connect.ts
+++ b/src/standard-live-connect.ts
@@ -2,7 +2,7 @@
 /* eslint-disable */
 import { PixelSender } from './pixel/sender'
 import * as errorHandler from './events/error-pixel'
-import * as C from './utils/consts'
+import { getGlobalBus, LI_TOPICS } from 'live-connect-common'
 import { StateWrapper } from './pixel/state'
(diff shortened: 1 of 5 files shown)


src/standard-live-connect.ts
```diff
@@ -98,8 +96,8 @@ function _standardInitialization (liveConnectConfig: LiveConnectConfig, external
     const postManagedState = managers.reduce(reducer, enrichedState)
 
     const syncContainerData = mergeObjects(configWithPrivacy, { peopleVerifiedId: postManagedState.data.peopleVerifiedId })
-    const onPixelLoad = () => eventBus.emit(C.PIXEL_SENT_PREFIX, syncContainerData)
-    const onPixelPreload = () => eventBus.emit(C.PRELOAD_PIXEL, '0')
+    const onPixelLoad = () => getGlobalBus().emit(LI_TOPICS.PIXEL_SENT, syncContainerData)
```
mschuwalow, line 99: we can also emit to the local bus if we want to keep backwards compat

PR #115, opened 2023-05-23
CM-807: Upgrading build dependencies

SuperIzya:
All npm dependencies upgraded to remove outdated/deprecated/vulnerable ones.

Author Todo List:

- [x] Build in CI passes
- [x] Self-Review


package.json
```diff
@@ -49,85 +50,70 @@
     "release:ci:major": "release-it major --ci"
   },
   "dependencies": {
-    "live-connect-common": "^1.0.0",
+    "live-connect-common": "file:../live-connect-common/live-connect-common-3.0.0.tgz",
+    "live-connect-handlers": "file:../live-connect-handlers/live-connect-handlers-3.0.0.tgz",
```
SuperIzya, line 54: Will be set to a proper dependencies when `live-connect-common` and `live-connect-handlers` are published.

rollup/dist.config.js
```diff
@@ -1,28 +1,36 @@
...
 export default {
-  input: './src/index.ts',
+  ...config,
   output: [
     {
-      file: `${OUTPUT_DIR}/index.js`,
-      format: 'cjs'
+      file: `${OUTPUT_DIR}/index.cjs`,
+      format: 'cjs',
+      sourcemap: false
```
SuperIzya, line 18: Turn the source map on for `live-connect`. Make sure it is off for `liveconnectbuilder`.

src/standard-live-connect.ts
```diff
@@ -119,23 +119,23 @@ function _standardInitialization (liveConnectConfig: LiveConnectConfig, external
   }
 }
 
-function _initializeWithoutGlobalName(liveConnectConfig: LiveConnectConfig, externalStorageHandler: ExternalStorageHandler, externalCallHandler: ExternalCallHandler, eventBus: EventBus): ILiveConnect {
+function _initializeWithoutGlobalName(liveConnectConfig: LiveConnectConfig, externalStorageHandler: StorageHandler, externalCallHandler: ExternalCallHandler, eventBus: EventBus): ILiveConnect {
```
SuperIzya, line 122: ExternalCallHandler? WTF?

.circleci/config.yml
```diff
@@ -83,8 +83,11 @@ jobs:
           path: ./test-results
           destination: test-results
       - run:
-          name: Code coverage
-          command: npm run coverage
+          name: Code Coverage
+          command: |
+            curl -Os https://uploader.codecov.io/latest/linux/codecov
+            chmod +x codecov
+            ./codecov -t ${CODECOV_TOKEN}
```
wi101, line 90: I think the value of CODECOV_TOKEN needs to be set in circleci env vars, or does it read it from somewhere else?

rollup/test-resources.config.js
```diff
@@ -1,13 +1,9 @@
...
 import replace from '@rollup/plugin-replace'
-import packageJson from '../package.json'
-import cleaner from 'rollup-plugin-cleaner'
-import strip from '@rollup/plugin-strip'
-import ts from "rollup-plugin-ts";
+import fs from 'fs'
+import path from 'path'
+import config from './dist.config.js'
+
+const packageJson = JSON.parse(fs.readFileSync(path.resolve('package.json'), { encoding: 'utf-8' }))
```
3link, line 6: Wouldn't `import packageJson from '../package.json' assert { type: "json" }` be more concise? You could also get rid of the `fs` and `path` imports.

src/standard-live-connect.ts
3link: Good questions.

src/utils/page.ts
```diff
@@ -55,9 +55,7 @@ function _collectElementsText(contextSelectors: string, contextElementsLength: n
 
 function encodedByteCount(s: string): number {
   // From: https://stackoverflow.com/questions/2219526/how-many-bytes-in-a-javascript-string
-  const utf8Bytelength = encodeURI(s).split(/%..|./).length - 1
-  const base64EncodedLength = 4 * Math.ceil(utf8Bytelength / 3.0)
-  return base64EncodedLength
+  return new Blob([s]).size
```
3link, line 58: It changes the semantics of this function: it first counted how many bytes are in the passed string and then computed the length of base64 encoding of these bytes with is `ceil(4/3*n)`. With this implementation, the script will send ~ `1/3` more data than configured.

test/it/live-connect.spec.ts
```diff
@@ -13,9 +15,14 @@ import {
...
-} from './helpers/browser'
+} from './helpers/browser.ts'
 import dirtyChai from 'dirty-chai'
-import packageJson from '../../package.json'
+import fs from 'fs'
+import path from 'path'
+import { fileURLToPath } from 'url'
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url))
+const packageJson = JSON.parse(fs.readFileSync(path.resolve(__dirname, '..', '..', 'package.json'), { encoding: 'utf-8' }))
```
3link, line 25: This could be an import too.

test/unit/utils/page.spec.ts
```diff
@@ -42,107 +52,162 @@ describe('Page Utils', () => {
...
+      ...iframe.contentWindow,
+      parent: win
+    }
+
+    // eslint-disable-next-line no-global-assign
+    window = win
+
+    expect(getPage(fw)).to.be.eql('https://liveintent.com/about?key=value')
+    // eslint-disable-next-line no-global-assign
+    window = oldWindow
```
3link, line 76: Should a try with resources that guarantees that the window is restored to its old state be used here? Something like this: https://github.com/LiveIntent/live-connect/pull/116/files#diff-b1c4e1833e06200917cbcbfcd57a775db133813709cf27170f2325bc45a4bcf9?

test/it/live-connect.spec.ts
SuperIzya: See above

.circleci/config.yml
wi101: You can get the CODECOV_TOKEN in circleci env variables here: https://app.codecov.io/gh/LiveIntent/live-connect/settings

src/standard-live-connect.ts
wi101: Good catch, consider to change it ExternalCallHandler -> CallHandler like you did in some other places.

tsconfig.json
```diff
@@ -1,13 +1,17 @@
...
-    "types": ["webdriverio/async", "mocha"],
+    "types": ["mocha"],
     "strict": true,
-    "declaration": true
+    "declaration": true,
+    "sourceMap": true,
+    "moduleResolution": "Node",
+    "outDir": "./dist",
+    "module": "ESNext",
+    "target": "es5"
```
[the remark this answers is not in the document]
SuperIzya: Old browsers (e.g. Chrome 65) understand only `es5` unfortunately.

tsconfig-test.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "module": "CommonJS"
```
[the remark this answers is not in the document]
SuperIzya: `ts-mocha` doesn't work with `ESNext`, whereas `@rollup/plugin-ts` can't optimize modules other than `ESNext`.

test/unit/utils/page.spec.ts
SuperIzya: This is a test. If it fails, then the code is wrong. I actually want it to fail in most spectacular way. That's why I don't want to put `try-finally` in this particular case.

rollup/test-resources.config.js
3link: What do you mean? What are the situations it does not work?

test/unit/utils/page.spec.ts
3link: Are the changes to `window` visible outside of that test -e.g. if the expectation is not fulfilled and the test fails? If yes, I would rather prefer to keep it local to that test.

SuperIzya: Again, **if** the `expect` fails and the code never gets to `window = oldWindow`, which will lead to other tests failing - I don't really mind. IMHO, the tests should work in optimistic scenario and fail in any other way.

3link: I think tests should be isolated so that it is easy to understand what is broken. A test having effect on other tests - even if it is only in case of failure - is usually not what you want.

rollup/test-resources.config.js
3link: Bumping it ⬆️

test/unit/utils/page.spec.ts
SuperIzya: Ok. Will fix that....

SuperIzya: Actually, it does not affect other tests.
Tested with failing expectation `expect(getPage(fw)).to.be.eql('https://liveintent.com/about1?key=value')` - only 1 failed test

rollup/test-resources.config.js
SuperIzya: ` ExperimentalWarning: Import assertions are not a stable feature of the JavaScript language. Avoid relying on their current behavior and syntax as those might change in a future version of Node.js.`

3link: 👍

test/unit/utils/page.spec.ts
3link: My point is that if it is used by other tests, it should be restored to its original state before the test ends. Tests should have no effect on each other and should not run using each other's state. I do not want to discuss it for too long. I leave it up to you if you change it or not.

3link (approved):

mschuwalow (approved):

PR #118, opened 2023-06-01
NO ISSUE - Remove daily v3 builds

3link:
There no need to daily build the v3 branch.

Author Todo List:

- [x] Build in CI passes
- [x] Self-Review

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,11 +3,6 @@ version: 2.1
 orbs:
   node: circleci/node@5.0.3
 
-constants:
-  releasable_branches: &releasable_branches
-    - master
-    - v3.x
-
 parameters:
   node_version:
     type: string
@@ -153,7 +148,7 @@ workflows:
             - browserstack
           filters:
             branches:
-              only: *releasable_branches
+              only: master
 
       - release:
           context: live-connect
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -153,7 +148,7 @@ workflows:
             - browserstack
           filters:
             branches:
-              only: *releasable_branches
+              only: master
```
SuperIzya, line 151: FYI: This will be overwritten by https://github.com/LiveIntent/live-connect/pull/115. IMHO, it's better to leave alias, since we already have it.

3link: I think without the `releasable_branches` constant it might actually be cleaner. We can not use the constant for `release_major?` since it is only reserved for `master`. Thus in `release_major?` we would still use `master` and in all other steps we would use `releasable_branches` which for now only consist of `master`. That is confusing. I think we should re-introduce `releasable_branches` when really required. 

> FYI: This will be overwritten by https://github.com/LiveIntent/live-connect/pull/115.
Please do no overwrite everything. It needs to be merged properly.

SuperIzya: I will merge it manually and submit to review, yours truly CO

SuperIzya (approved):

PR #182, opened 2023-09-07
[CM-940] Non-zero default timeout for ajax

mschuwalow:
# [CM-940](https://liveintent.atlassian.net/browse/CM-940)

Author Todo List:

- [x] Build in CI passes
- [x] Self-Review

[CM-940]: https://liveintent.atlassian.net/browse/CM-940?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pixel/sender.ts
+++ b/src/pixel/sender.ts
@@ -4,7 +4,7 @@ import { LiveConnectConfig, EventBus } from '../types'
 import { StateWrapper } from './state'
 import { WrappedCallHandler } from '../handlers/call-handler'
 
-const DEFAULT_AJAX_TIMEOUT = 0
+const DEFAULT_AJAX_TIMEOUT = 5000
 
 export class PixelSender {
   url: string


src/pixel/sender.ts
```diff
@@ -4,7 +4,7 @@ import { LiveConnectConfig, EventBus } from '../types'
 import { StateWrapper } from './state'
 import { WrappedCallHandler } from '../handlers/call-handler'
 
-const DEFAULT_AJAX_TIMEOUT = 0
+const DEFAULT_AJAX_TIMEOUT = 5000
```
wi101, line 7: why?
I thought 0 means there is no timeout 
https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/timeout

mschuwalow: We see error logs that suggest that some browser timeout with a timeout of 0.
This is an attempt to make these go away

PR #276, opened 2023-12-06
CM-1047: Fix live-connect flaky browserstack test

wi101:
Investigate why Chrome 40 and IE tests are failing.
`array#includes` is not supported by IE and Chrome40
we can either:
🔴  use the polyfill but this will increase the size of the code.
🟢 replace includes with a supported function (indexOf). the code size remains the same.

Author Todo List:

- [x] Build in CI passes
- [x] Self-Review

### DIFF
--- a/test-config/wdio.browserstack.conf.mjs
+++ b/test-config/wdio.browserstack.conf.mjs
@@ -13,9 +13,10 @@ const commonBStackCapabilities = {
 // https://www.browserstack.com/automate/capabilities
 const allCapabilities = [
   { browserName: 'Chrome', browserVersion: 'latest', 'bstack:options': { ...commonBStackCapabilities, os: 'Windows', osVersion: '10' } },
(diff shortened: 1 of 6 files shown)


test-config/wdio.browserstack.conf.mjs
```diff
@@ -13,9 +13,10 @@ const commonBStackCapabilities = {
 // https://www.browserstack.com/automate/capabilities
 const allCapabilities = [
   { browserName: 'Chrome', browserVersion: 'latest', 'bstack:options': { ...commonBStackCapabilities, os: 'Windows', osVersion: '10' } },
-  { browserName: 'Chrome', browserVersion: '47.0', 'bstack:options': { ...commonBStackCapabilities, os: 'Windows', osVersion: '7' } },
+  { browserName: 'Chrome', browserVersion: '40.0', 'bstack:options': { ...commonBStackCapabilities, os: 'Windows', osVersion: '7' } },
```
SuperIzya, line 16: Chrome version can be safely bumped to `45`.

PR #289, opened 2023-12-21
CM-971: Rename trackerName to trackerVersion

wi101:
The trackerName is confusing.
This field will be renamed in PixelEvent as well.

Author Todo List:

- [ ] Build in CI passes
- [ ] Self-Review

### DIFF
--- a/COLLECTOR_PARAMS.md
+++ b/COLLECTOR_PARAMS.md
@@ -8,8 +8,8 @@
 - contains the b64 url encoded string of the JSON that was sent via `liveConnect.push` function
 ### `duid`
 - contains the LiveConnect managed first party identifier, in the `${apexDomainHash}--${ULID}`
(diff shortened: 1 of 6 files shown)


SuperIzya (approved):

COLLECTOR_PARAMS.md
```diff
@@ -8,8 +8,8 @@
 - contains the b64 url encoded string of the JSON that was sent via `liveConnect.push` function
 ### `duid`
 - contains the LiveConnect managed first party identifier, in the `${apexDomainHash}--${ULID}`
-### `tna`
```
mschuwalow, line 11: I'm fine with the change. But given that we need to maintain backwards compat, I'm not sure this will be any less confusing

mschuwalow (approved):

PR #302, opened 2024-01-09
CM-971: consider trackerName field for backward compatibility 

wi101:
If we receive trackerName from older LC versions or through prebid, it will be forwarded as `tv` to LP.

Author Todo List:

- [ ] Build in CI passes
- [ ] Self-Review

### DIFF
--- a/src/pixel/state.ts
+++ b/src/pixel/state.ts
@@ -26,6 +26,8 @@ const paramExtractors: ((state: State) => [string, string][])[] = [
   ifDefined('eventSource', source => asParamOrEmpty('se', source, (s) => base64UrlEncode(JSON.stringify(s, replacer)))),
   ifDefined('liveConnectId', fpc => asStringParam('duid', fpc)),
   ifDefined('trackerVersion', v => asStringParam('tv', v)),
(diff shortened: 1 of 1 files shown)


src/pixel/state.ts
```diff
@@ -26,6 +26,9 @@ const paramExtractors: ((state: State) => [string, string][])[] = [
   ifDefined('eventSource', source => asParamOrEmpty('se', source, (s) => base64UrlEncode(JSON.stringify(s, replacer)))),
   ifDefined('liveConnectId', fpc => asStringParam('duid', fpc)),
   ifDefined('trackerVersion', v => asStringParam('tv', v)),
+  // For backward compatibility
+  // @ts-ignore
+  ifDefined('trackerName', v => asStringParam('tv', v)),
```
3link, line 31: In what cases is this required?

wi101: thanks for asking
I misinterpreted a scenario related to prebid old versions, but we won't run into that case as the old dependency would be imported..

wi101:
Not needed.

PR #348, opened 2024-03-06
[No Ticket] Adjust babel and test options

mschuwalow:
Author Todo List:

- [x] Build in CI passes
- [x] Self-Review


3link (approved):

peixunzhang (approved):

rollup/dist.config.js
```diff
@@ -13,9 +12,10 @@ export default [
         input: ['./src/index.ts', './src/internal.ts'],
         output: [
             {
-                dir: `${OUTPUT_DIR}`,
-                format: 'cjs',
-                sourcemap: true
```
mschuwalow, line 18: Sourcemaps didn't work for me properly when debugging in vscode. I propose we just don't terse the code, then no special handling is necessary

rollup/dist.config.js
```diff
@@ -13,9 +12,10 @@ export default [
...
+                dir: OUTPUT_DIR,
+                entryFileNames: '[name].cjs',
+                chunkFileNames: '[name]-[hash].cjs',
```
SuperIzya, line 17: Why chunks? Are you splitting it?

src/cache.ts
```diff
@@ -1,5 +1,5 @@
 import { EventBus, isObject, expiresInDays } from 'live-connect-common'
-import { WrappedStorageHandler } from './handlers/storage-handler'
+import { WrappedStorageHandler } from './handlers/storage-handler.js'
```
SuperIzya, line 2: Importing local `ts` files as `js` files? Wut?

rollup/dist.config.js
SuperIzya: Did you checked chrome's dev-tools? Do you see sources there?

rollup/dist.config.js
mschuwalow: This is internally used by rollup. By default it will generate them with plain .js extension, but if we switch to module, we need to give them explicit .cjs extension (as they are commonjs).

Example:
```
❯ la dist
Permissions Size User       Group Date Modified Git Name
.rw-r--r--   585 mschuwalow staff  6 Mar 17:09   -I index.cjs
.rw-r--r--   276 mschuwalow staff  6 Mar 17:09   -I index.cjs.map
.rw-r--r--   307 mschuwalow staff  6 Mar 17:09   -I index.cjs.mjs
.rw-r--r--  2.9k mschuwalow staff  6 Mar 17:09   -I index.d.ts
.rw-r--r--   66k mschuwalow staff  6 Mar 17:09   -I initializer-me88Mdta.cjs
.rw-r--r--   64k mschuwalow staff  6 Mar 17:09   -I initializer-me88Mdta.cjs.map
.rw-r--r--   584 mschuwalow staff  6 Mar 17:09   -I internal.cjs
.rw-r--r--   107 mschuwalow staff  6 Mar 17:09   -I internal.cjs.map
.rw-r--r--   470 mschuwalow staff  6 Mar 17:09   -I internal.cjs.mjs
.rw-r--r--  1.1k mschuwalow staff  6 Mar 17:09   -I internal.d.ts
.rw-r--r--  6.6k mschuwalow staff  6 Mar 17:09   -I standard-live-connect.d-CGhFVSFd.d.ts
```

src/cache.ts
mschuwalow: Normal node resolution mode. Esm modules want to have an explicit extension.
This can be disabled with --experimental-specifier-resolution=node, but I decided to just stick with the defaults

mschuwalow:
`Did you checked chrome's dev-tools? Do you see sources there?`

For this to work we would need to emit sourcemaps from liveconnectbuilder, which we are not. Here it's really only relevant when getting used as a library.

src/cache.ts
SuperIzya: This is wired and contr-intuitive. Can you at least make it consume `ts` extension? I'm not sure the navigation will work in IntelliJ as it is right now...

mschuwalow: See here for more details for example https://stackoverflow.com/questions/62619058/appending-js-extension-on-relative-import-statements-during-typescript-compilat.

test/unit/minimal-live-connect.spec.ts
```diff
@@ -3,10 +3,10 @@
 import jsdom from 'global-jsdom'
 import sinon from 'sinon'
 import { expect, use } from 'chai'
-import { MinimalLiveConnect, LiveConnect } from '../../src'
+import { MinimalLiveConnect, LiveConnect } from '../../src/index.js'
```
SuperIzya, line 6: Node resolution can't find `index` automatically?

src/cache.ts
mschuwalow: Can you try navigation out? It should work without any issues (as this is pretty much the 'default' setup now)

test/unit/minimal-live-connect.spec.ts
mschuwalow: that should work, will put them back.

mschuwalow: nope https://stackoverflow.com/a/68621282

PR #352, opened 2024-03-11
CM-1181 - Send page URL to the identity resolution endpoint

3link:
Send the page URL to the identity resolution endpoint. This is helpful in cases where the origin header is not available. 

Author Todo List:

- [x] Build in CI passes
- [x] Self-Review


src/pixel/url-collector.ts
```diff
@@ -29,6 +29,17 @@ export function collectUrl(state: UrlState): [string, boolean, string[]] {
   return [url.toString(), pathRemoved, blockedParams]
 }
 
+export function stripQueryAndPath(pageUrl?: string): string {
```
mschuwalow, line 32: Empty string will be stripped by the query builder default. In case this is the desired behaviour, I propose more explicitly communicating that in the types by returning undefined

src/pixel/url-collector.ts
```diff
@@ -29,6 +29,17 @@ export function collectUrl(state: UrlState): [string, boolean, string[]] {
...
+  if (pageUrl === undefined || pageUrl.length === 0) {
+    return ''
+  } else {
+    const url = new ParsedUrl(pageUrl)
+    url.pathname = ''
+    url.search = ''
+    return url.toString()
+  }
+}
```
mschuwalow, line 41:
```suggestion
export function stripQueryAndPath(pageUrl?: string): string | undefined {
  if (pageUrl === undefined || pageUrl.length === 0) {
    return undefined
  } else {
    const url = new ParsedUrl(pageUrl)
    url.pathname = ''
    url.search = ''
    return url.toString()
  }
}
```

mschuwalow (approved):

src/idex.ts
```diff
@@ -59,6 +62,7 @@ export class IdentityResolver {
       .addOptional('gpp_as', nonNullConfig.gppApplicableSections?.join(','))
       .addOptional('cd', nonNullConfig.cookieDomain)
       .addOptional('ic', encodeIdCookie(nonNullConfig.resolvedIdCookie), { stripEmpty: false })
+      .addOptional('pu', stripQueryAndPath(nonNullConfig.pageUrl))
```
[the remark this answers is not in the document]
3link: Nice. Added that.

src/pixel/url-collector.ts
3link: Yes, desired behaviour. Changed the signature to `stripQueryAndPath(pageUrl: string): string` and the impl uses `onNonNull` now.

wi101 (approved):

peixunzhang (approved):

